Issue #140, opened 2026-04-20
[FEATURE] support setting client host, client port and client secret for Oauth flows


akaralar:
## Problem

Slack and Figma have specific needs: Figma requires the host to be "localhost" instead of ip address, Slack requires a specific port and a specific callback path.

## Proposed Solution

Provide fields that can be filled in the yaml file, and fallback to defaults if there is none. 

## Additional Context

I have added these to my fork at https://github.com/akaralar/mcp-gateway/tree/fix/wire-oauth-client-id-from-config and it works fine as far as I can tell, I can isolate the changes and open a PR if you think it makes sense to have these. Just so you know, I made my changes with Claude Code so it's AI assisted.

MikkoParkkola:
Thanks for the detailed writeup, @akaralar — and bonus points for already sketching the fix on your fork. This is a real gap: the callback server today is hardcoded to `http://127.0.0.1:<random-port>/oauth/callback` (see [`src/oauth/callback.rs`](https://github.com/MikkoParkkola/mcp-gateway/blob/main/src/oauth/callback.rs#L84-L98)) and `OAuthConfig` (in [`src/config/mod.rs`](https://github.com/MikkoParkkola/mcp-gateway/blob/main/src/config/mod.rs#L520-L535)) only exposes `enabled`, `scopes`, `client_id`, and `token_refresh_buffer_secs` — no host, no port, no secret. So the three providers you mentioned really can't complete a flow against the current gateway.

**Proposed shape** (matches what's on your fork, with a couple of safety nits):

```yaml
backends:
  slack:
    transport: { http_url: "https://slack.com/api/mcp" }
    oauth:
      enabled: true
      client_id: "${SLACK_CLIENT_ID}"
      client_secret: "${SLACK_CLIENT_SECRET}"   # NEW — env-expanded, never logged
      callback_host: "localhost"                # NEW — default "127.0.0.1"
      callback_port: 3000                       # NEW — default: OS-picked ephemeral
      callback_path: "/slack/oauth_redirect"    # NEW — default "/oauth/callback"
      scopes: ["chat:write"]
```

Concretely in `OAuthConfig`:
- `client_secret: Option<String>` — passed as `client_secret` in the token exchange when present (confidential clients). Must be env-expanded via the existing `expand_env` pass on load (your fork already does this — good); must be excluded from `Debug`/`Display` and from any config-dump endpoint; should preferentially come from `${ENV}` references, with a warning if a literal string is used.
[...]

PR #141, opened 2026-04-20
feat(oauth): configurable client_id, client_secret, callback host/port/path (#140)

MikkoParkkola:
Based on @akaralar's fork (`akaralar/mcp-gateway@fix/wire-oauth-client-id-from-config`). Thanks for the contribution — this closes the gap flagged in #140 where Slack and Figma can't complete an OAuth flow against the gateway.

## Summary

Adds four new optional fields to `OAuthConfig`:

- `client_id: Option<String>` — pre-configured client ID (skips dynamic registration when set)
- `client_secret: Option<String>` — sent as `client_secret` on token exchange when present (confidential clients)
- `callback_host: Option<String>` — default `"127.0.0.1"`, needed as `"localhost"` for Figma
- `callback_port: Option<u16>` — fixed callback port (Slack requires a specific port)
- `callback_path: Option<String>` — default `"/oauth/callback"`, Slack uses `/slack/oauth_redirect`

Wires them through `Backend::build_oauth_client` -> `OAuthClient::new` -> `callback::start_callback_server`, and the token exchange in `OAuthClient`. `client_id` and `client_secret` are passed through the existing `expand_env` pass on config load so `${SLACK_CLIENT_SECRET}` references are resolved.

Also: if dynamic registration returns a `client_secret`, it is captured and used on subsequent token exchanges.

## Clusters evaluated from the fork

The fork had two parallel clusters on `fix/wire-oauth-client-id-from-config`:

1. **OAuth configurability** (commits `cc402def`, `fa412776`, `70963d0b`) — **this PR.**
2. **HTTP stale-session auto re-initialize** (commit `2b2a689c`) — deliberately not included here. Different topic, and the silent auto-reinit deserves maintainer design input (see comment on #140).

(The earlier triage that mentioned A2A transport and CI hash-pinning was incorrect — both are already in upstream; fork had only the four OAuth commits.)

## Tweaks applied on top of akaralar's commits

One extra commit `chore(oauth): rustfmt + clippy allow` (co-authored to @akaralar):

- `cargo fmt` differences in `src/oauth/callback.rs` and `src/oauth/client/mod.rs`.
- `#[allow(clippy::too_many_arguments)]` on `OAuthClient::new` — it now takes 11 inputs, above clippy's pedantic limit of 7. A natural follow-up is to bundle these into an `OAuthClientConfig` struct; leaving that to a focused refactor.

## Open design questions (not blocking this PR, but flagged from issue #140)

@MikkoParkkola raised two in the issue thread that this PR does **not** yet address:

1. **`client_secret` literal vs `\${ENV}`-only** — currently accepts both. Should emit a warning (or reject) literal values?
2. **`callback_host` whitelist** — currently accepts any string. Mikko suggested whitelisting to `{\"127.0.0.1\", \"localhost\"}` to avoid SSRF / open-redirect weirdness on the local callback server.
3. **`callback_path` validation** — should enforce starts-with-`/`, no query string, no `..`.
4. **PKCE + `client_secret` combo** — some providers require both, some only one. Worth confirming against Slack/Figma docs.

Happy to tighten these in a follow-up commit on this branch if you want them in before merge, or land as a separate hardening PR.

## Verification

Run on an Apple-Silicon mac, on the `feat/oauth-configurability` branch (4 commits ahead of `main` after `625738fe`):

- `cargo build --all-targets` -> finished in 4m 14s, no errors
- `cargo test --all-features --lib` -> **2672 passed, 0 failed**
- `cargo test --all-features --lib --bins` -> 111 additional bin tests passed
- `cargo clippy --all-targets --all-features -- -D warnings` -> clean
- `cargo fmt --check` -> clean

## Authorship

Three commits retain `Ahmet Karalar <ahmet.karalar@backmarket.com>` as author; the fmt/clippy commit is co-authored to him.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -529,6 +529,9 @@ pub struct OAuthConfig {
     /// Client ID (optional — uses dynamic registration or generates one if not set).
     #[serde(default)]
     pub client_id: Option<String>,
+    /// Client secret (optional — required by some providers for token exchange).
+    #[serde(default)]
+    pub client_secret: Option<String>,
     /// Seconds before expiry to proactively refresh the token (default: 300).
     #[serde(default = "default_token_refresh_buffer")]
     pub token_refresh_buffer_secs: u64,
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -540,6 +540,18 @@ pub struct OAuthConfig {
     /// Client secret (optional — required by some providers for token exchange).
     #[serde(default)]
     pub client_secret: Option<String>,
+    /// Fixed port for the OAuth callback server (optional — uses a random port if not set).
+    /// Some providers (e.g., Slack) require the callback on a specific port.
+    #[serde(default)]
+    pub callback_port: Option<u16>,
+    /// Callback path for the OAuth redirect (default: "/oauth/callback").
+    /// Some providers register a specific path (e.g., "/callback").
+    #[serde(default)]
+    pub callback_path: Option<String>,
+    /// Callback host for the OAuth redirect URI (default: "127.0.0.1").
+    /// Some providers register "localhost" instead of "127.0.0.1".
+    #[serde(default)]
+    pub callback_host: Option<String>,
     /// Seconds before expiry to proactively refresh the token (default: 300).
     #[serde(default = "default_token_refresh_buffer")]
     pub token_refresh_buffer_secs: u64,
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -246,6 +246,14 @@ impl Config {
             for value in backend.env.values_mut() {
                 *value = Self::expand_string(&re, value);
             }
+            if let Some(ref mut oauth) = backend.oauth {
+                if let Some(ref mut id) = oauth.client_id {
+                    *id = Self::expand_string(&re, id);
+                }
+                if let Some(ref mut secret) = oauth.client_secret {
+                    *secret = Self::expand_string(&re, secret);
+                }
+            }
         }
 
         for dir in &mut self.capabilities.directories {
--- a/src/oauth/callback.rs
+++ b/src/oauth/callback.rs
@@ -101,7 +101,10 @@ pub async fn start_callback_server(
         .local_addr()
         .map_err(|e| Error::OAuth(format!("Failed to get callback server address: {e}")))?;
 
-    let callback_url = format!("http://{callback_host}:{}{callback_path}", actual_addr.port());
+    let callback_url = format!(
+        "http://{callback_host}:{}{callback_path}",
+        actual_addr.port()
+    );
     info!(url = %callback_url, "OAuth callback server listening");
 
     // Create oneshot channel for the result
@@ -265,7 +268,9 @@ mod tests {
     #[tokio::test]
     async fn callback_flow_success() {
         let state = "csrf_state_123".to_string();
-        let server = start_callback_server(state.clone(), None, None, None).await.unwrap();
+        let server = start_callback_server(state.clone(), None, None, None)
+            .await
+            .unwrap();
         let callback_url = server.callback_url.clone();
 
         // Simulate the OAuth provider redirecting back
--- a/src/oauth/callback.rs
+++ b/src/oauth/callback.rs
@@ -84,10 +84,16 @@ impl CallbackServer {
 pub async fn start_callback_server(
     expected_state: String,
     port: Option<u16>,
+    path: Option<&str>,
+    host: Option<&str>,
 ) -> Result<CallbackServer> {
-    // Find an available port
-    let addr: SocketAddr = format!("127.0.0.1:{}", port.unwrap_or(0)).parse().unwrap();
-    let listener = TcpListener::bind(addr)
+    let callback_host = host.unwrap_or("127.0.0.1");
+    let callback_path = path.unwrap_or("/oauth/callback");
+
+    // Bind to 127.0.0.1 (even if callback_host is "localhost") since
+    // SocketAddr requires a numeric IP, but use callback_host in the URL.
+    let bind_addr: SocketAddr = format!("127.0.0.1:{}", port.unwrap_or(0)).parse().unwrap();
+    let listener = TcpListener::bind(bind_addr)
         .await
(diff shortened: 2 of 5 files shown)


MikkoParkkola (commented):
@akaralar welcome, and thank you for this. The Slack and Figma OAuth gap was a real one, and it's great to have someone dig into the fork and send a proper proposal back.

Left four inline suggestions, one-click apply. They bring the PR up to the bar we hold for OAuth code.

No pressure on these. If you want to land them yourself the PR stays fully yours, which would be lovely. If it's easier, just say so and I'll push them co-authored to you. You get full credit either way.

Filing three follow-up issues for stuff that doesn't block this merge: OAUTH_CONFIG.md, callback telemetry, IPv6 localhost. You're credited on those too.

Looking forward to having you in CONTRIBUTORS.md. We love seeing contributions like this.

Mikko

src/oauth/callback.rs
```diff
@@ -84,18 +84,27 @@ impl CallbackServer {
...
     expected_state: String,
     port: Option<u16>,
+    path: Option<&str>,
+    host: Option<&str>,
 ) -> Result<CallbackServer> {
-    // Find an available port
-    let addr: SocketAddr = format!("127.0.0.1:{}", port.unwrap_or(0)).parse().unwrap();
-    let listener = TcpListener::bind(addr)
+    let callback_host = host.unwrap_or("127.0.0.1");
+    let callback_path = path.unwrap_or("/oauth/callback");
```
MikkoParkkola, line 91: Default-deny whitelist on `callback_host`. Since this advertises a URL to the OAuth provider, an explicit allow-list closes a class of weird config-injection scenarios without constraining real usage. Rejecting at bind time gives a clearer error than a 404 at callback time.

```suggestion
    let callback_host = host.unwrap_or("127.0.0.1");
    let callback_path = path.unwrap_or("/oauth/callback");

    const ALLOWED_CALLBACK_HOSTS: &[&str] = &["127.0.0.1", "localhost"];
    if !ALLOWED_CALLBACK_HOSTS.contains(&callback_host) {
        return Err(Error::OAuth(format!(
            "callback_host {callback_host:?} not allowed; use one of {ALLOWED_CALLBACK_HOSTS:?} or open an issue if you need another"
        )));
    }
    if !callback_path.starts_with('/') || callback_path.contains("..") || callback_path.contains('?') || callback_path.contains('#') {
        return Err(Error::OAuth(format!(
            "callback_path {callback_path:?} must start with '/' and contain no '..', '?', or '#'"
        )));
    }
```

src/oauth/client/mod.rs
```diff
@@ -80,13 +92,19 @@ struct ClientRegistrationResponse {
...
         resource_url: String,
         scopes: Vec<String>,
         storage: Arc<TokenStorage>,
         token_refresh_buffer_secs: u64,
+        client_id: Option<String>,
+        client_secret: Option<String>,
+        callback_port: Option<u16>,
+        callback_path: Option<String>,
+        callback_host: Option<String>,
     ) -> Self {
```
MikkoParkkola, line 108: Pulling the 11 inputs into a config struct. Removes the `too_many_arguments` allow and keeps future additions clean (PKCE, per-provider scopes). Plain struct is enough, no builder needed.

```suggestion
    #[must_use]
    pub fn new(http_client: Client, config: OAuthClientConfig) -> Self {
```

With `OAuthClientConfig` defined near the top of the file:

```rust
pub struct OAuthClientConfig {
    pub backend_name: String,
    pub resource_url: String,
    pub scopes: Vec<String>,
    pub storage: Arc<TokenStorage>,
    pub token_refresh_buffer_secs: u64,
    pub client_id: Option<String>,
    pub client_secret: Option<String>,
    pub callback_port: Option<u16>,
    pub callback_path: Option<String>,
    pub callback_host: Option<String>,
}
```

Callsites in `src/backend/mod.rs` and the tests file will need the same shape.

src/oauth/client/tests.rs
```diff
@@ -117,6 +117,11 @@ fn new_client_has_no_valid_token() {
         vec!["read".to_string()],
         storage,
         300,
+        None,
+        None,
+        None,
+        None,
+        None,
```
MikkoParkkola, line 124: The `None × 5` keeps regression coverage but doesn't exercise the new paths. Three behavior tests cover it, same `wiremock` pattern used elsewhere in this file:

1. `callback_host_renders_in_url` — `host = Some("localhost")`, assert `callback_url` starts with `http://localhost:` while the listener still binds `127.0.0.1` numerically
2. `custom_callback_path_is_registered` — `path = Some("/slack/oauth_redirect")`, assert GET to that path hits the handler
3. `client_secret_included_in_token_exchange` — mock server asserts `client_secret` is present in the form body

src/config/mod.rs
```diff
@@ -246,6 +246,14 @@ impl Config {
             for value in backend.env.values_mut() {
                 *value = Self::expand_string(&re, value);
             }
+            if let Some(ref mut oauth) = backend.oauth {
+                if let Some(ref mut id) = oauth.client_id {
+                    *id = Self::expand_string(&re, id);
+                }
+                if let Some(ref mut secret) = oauth.client_secret {
+                    *secret = Self::expand_string(&re, secret);
+                }
```
MikkoParkkola, line 255: Warn when `client_secret` is a literal (not an env var). People paste secrets into config without realizing, a one-line log nudges them without rejecting the config outright.

```suggestion
                if let Some(ref mut secret) = oauth.client_secret {
                    if !secret.contains("${") && secret.len() > 8 {
                        tracing::warn!(
                            backend = %backend.name,
                            "OAuth client_secret appears to be a literal; prefer ${{ENV_VAR}} for secrets"
                        );
                    }
                    *secret = Self::expand_string(&re, secret);
                }
```

akaralar:
@MikkoParkkola sorry I am just seeing all these notifications, I can take a look over the weekend

MikkoParkkola:
Closing as superseded by the current main branch OAuth implementation. main now carries the configurable client_id/client_secret/callback host/port/path surface via OAuthClientConfig plus the newer localhost dual-bind callback handling, so this older branch conflicts with and would regress that implementation.

PR #159, opened 2026-04-27
fix: return structuredContent for tools with outputSchema

### LINKED ISSUE #158: gateway_search_tools declares outputSchema but wrap_tool_success never returns structuredContent
## Bug

gateway_search_tools is the only meta-tool that declares an outputSchema (via search_tools_output_schema() in src/gateway/meta_mcp_tool_defs.rs). However, the response path for all meta-tools goes through wrap_tool_success in src/gateway/meta_mcp_helpers.rs, which always builds a ToolsCallResult with content: [Content::Text { ... }] and never populates structuredContent.

Per the MCP spec (2025-06-18), when a tool advertises outputSchema, the response must include structuredContent. Clients that enforce this (e.g. the Python SDK mcp>=1.24.0, Kiro) reject the response with:

    MCP error -32600: Tool gateway_search_tools has an output schema but did not return structured content

All other meta-tools work fine because they set output_schema: None.

## Where

- Tool definition: src/gateway/meta_mcp_tool_defs.rs - build_search_tools_tool() sets output_schema: Some(search_tools_output_schema())
- Response builder: src/gateway/meta_mcp_helpers.rs - wrap_tool_success() always returns Content::Text, never structuredContent
- Dispatch: src/gateway/meta_mcp/mod.rs - handle_tools_call() routes all Ok(content) through wrap_tool_success

## Suggested Fix

Either:

1. Quick: Set output_schema: None on gateway_search_tools to match every other meta-tool.
2. Proper: Make wrap_tool_success schema-aware - when the tool has an outputSchema, populate structuredContent with the JSON value.

## Reproduction

Call gateway_search_tools from any MCP client that validates outputSchema/structuredContent compliance. For example, using the mcp Python SDK >= 1.24.0:

    result = await session.call_tool("gateway_search_tools", {"query": "list projects"})
    # Raises: McpError -32600

## Version

mcp-gateway 2.11.0

v4de:
## Summary

Fixes #158 - gateway_search_tools declares an outputSchema but the response never includes structuredContent, causing MCP clients that enforce the spec to reject it with -32600.

## Changes

- **src/protocol/messages.rs**: Added optional structured_content field to ToolsCallResult, serialized as structuredContent per the MCP spec (2025-06-18)
- **src/gateway/meta_mcp_helpers.rs**: Added has_output_schema parameter to wrap_tool_success. When true, populates structuredContent with the raw JSON value alongside the text content fallback
- **src/gateway/meta_mcp/mod.rs**: Updated dispatch to pass has_output_schema=true for gateway_search_tools (the only meta-tool with outputSchema)
- **src/gateway/meta_mcp_helpers_tests.rs**: Updated existing tests, added new test verifying structuredContent is present when has_output_schema is true

## Approach

Rather than removing the outputSchema from gateway_search_tools (which would lose the schema contract), this fix makes wrap_tool_success schema-aware. A simple boolean parameter controls whether structuredContent is populated. The text content is always included as a backward-compatible fallback.

This is forward-compatible: if more meta-tools gain outputSchema in the future, they just need to pass true at their call site.

### DIFF
--- a/src/gateway/meta_mcp/mod.rs
+++ b/src/gateway/meta_mcp/mod.rs
@@ -742,7 +742,7 @@ impl MetaMcp {
             return match result {
                 Ok(content) => {
                     use super::meta_mcp_helpers::wrap_tool_success;
-                    wrap_tool_success(id, &content)
+                    wrap_tool_success(id, &content, false)
                 }
                 Err(e) => JsonRpcResponse::error(Some(id), e.to_rpc_code(), e.to_string()),
             };
@@ -803,7 +803,10 @@ impl MetaMcp {
         };
 
         match result {
-            Ok(content) => wrap_tool_success(id, &content),
+            Ok(content) => {
+                let has_output_schema = tool_name == "gateway_search_tools";
+                wrap_tool_success(id, &content, has_output_schema)
+            }
             Err(e) => JsonRpcResponse::error(Some(id), e.to_rpc_code(), e.to_string()),
         }
     }
--- a/src/gateway/meta_mcp_helpers.rs
+++ b/src/gateway/meta_mcp_helpers.rs
@@ -650,12 +650,22 @@ pub(crate) fn build_circuit_breaker_stats_json(server: &str, stats: &CircuitBrea
 }
 
 /// Wrap a successful tool result `Value` into a `JsonRpcResponse`.
-pub(crate) fn wrap_tool_success(id: RequestId, content: &Value) -> JsonRpcResponse {
+///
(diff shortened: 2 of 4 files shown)


MikkoParkkola:
@v4de merged. thanks for this. broken structuredContent on tools with outputSchema means every spec-compliant client was silently rejecting gateway_search_tools responses. clean fix, good tests, no regressions. appreciate the report (#158) AND the PR. that's how it's done.

v4de:
@MikkoParkkola looks like the build failed, I am unable to build/test locally, was hoping the CI ran before the merge. Let me fix the missing field.

PR #256, opened 2026-06-16
fix(failsafe): auto-recover a tripped circuit breaker when a backend respawns

MikkoParkkola:
## Problem

A backend whose stdio child process dies and respawns leaves its circuit breaker wedged `Open` indefinitely. Observed live on the `trvl` backend: the breaker stayed open for >7 minutes while the child process itself was healthy and answering `ping` directly.

Root cause — the gateway has no working closed-loop recovery. The background health check could not heal a broken backend for three compounding reasons:

1. **Skipped down backends** — `is_running()` guard meant a dead backend (which reports `is_running()==false`) was never probed, exactly when it needed recovery.
2. **Pinged through the breaker** — `backend.request("ping")` short-circuits on `can_proceed()` and returns `CircuitOpen` without touching the backend, so the check could never discover the backend had recovered.
3. **Failure only logged** — no transport rebuild, no breaker reset.

Recovery was therefore left to organic traffic threading `success_threshold` (3) consecutive clean probes through 30s half-open windows, against an `is_connected()` that only read a cached `AtomicBool` (could report a zombie child as alive). With `reset_timeout` (30s) phase-locked to the health interval (30s), any single probe failure reset the counter, so it rarely converged.

## Changes

| # | Change | File |
|---|---|---|
| A | `Backend::health_probe()` — active recovery probe that **bypasses the breaker**, resets it on success, rebuilds the transport on failure. `force_restart()` rebuilds unconditionally (kill child + respawn + re-init). | `backend/mod.rs` |
| B | Health loop probes **running and breaker-tripped** backends; cleanly-idle backends left to the idle reaper. | `gateway/server/mod.rs` |
| C | `StdioTransport::is_connected()` confirms liveness via non-blocking `try_wait()` so a stale flag can't make `ensure_started()` a no-op. | `transport/stdio.rs` |
| D | Default health interval 30s → 10s, decoupled from the 30s breaker reset timeout (removes phase-lock; faster recovery). | `config/features/failsafe.rs` |
| E | `POST /ui/api/backends/{name}/revive` admin route — out-of-band operator lever (reset breaker + force rebuild) usable via curl, independent of the MCP tool surface. | `gateway/ui/backends.rs` |

A + B + C deliver automatic recovery within ~one health interval; D removes the phase-lock; E is the manual safety valve.

## Tests

- `health_probe_resets_tripped_breaker_on_success` (headline regression)
- `health_probe_failure_leaves_breaker_tripped`
- `is_circuit_tripped_reflects_breaker_state`
- `health_interval_is_decoupled_from_breaker_reset_timeout`
- `health_timeout_is_shorter_than_interval`

Full lib suite: 2847 pass. `cargo clippy --all-targets -- -D warnings` clean. `cargo fmt --check` clean.

## Scope / blast radius

5 files, +306/-5. New public API is additive (`is_circuit_tripped`, `force_restart`, `health_probe`, one HTTP route). No change to the Meta-MCP tool surface (E is an HTTP admin route, not a meta-tool).

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -843,6 +843,98 @@ impl Backend {
         self.failsafe.circuit_breaker.reset();
     }
 
+    /// Whether this backend's circuit breaker is currently tripped
+    /// (`Open` or `HalfOpen` — i.e. not `Closed`).
+    #[must_use]
+    pub fn is_circuit_tripped(&self) -> bool {
+        self.failsafe.circuit_breaker.state() != crate::failsafe::CircuitState::Closed
+    }
+
+    /// Tear down the current transport (killing any child process) and start a
+    /// fresh one.
+    ///
+    /// Unlike [`ensure_started`](Self::ensure_started), this does **not** trust
+    /// `is_connected()` — it always rebuilds. A wedged-but-not-exited child
+    /// (responds to `try_wait` as alive yet never answers requests) cannot be
+    /// recovered by `ensure_started` alone; this is the escape hatch the health
+    /// loop uses when a probe fails.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the fresh transport fails to start or initialize.
+    pub async fn force_restart(&self) -> Result<()> {
+        let _guard = self.start_lock.lock().await;
+        // Take the transport out and drop the RwLock write guard *before*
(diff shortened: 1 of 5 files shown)


MikkoParkkola:
Tracking: MIK-6150 — https://linear.app/parm/issue/MIK-6150

PR #259, opened 2026-06-16
feat(attestation): per-action signed attestation tokens (MIK-5223, B1-IDENT)

MikkoParkkola:
Rescued from the misrouted branch `symphony/MIK-5223/ccelite-anthropic` (the winning attempt). Brings the per-action attestation module into mcp-gateway.

## What
Self-contained `src/attestation/` module (~1716 LOC) plus AC test suite:
- **signer** HMAC-SHA256 token signer, bnaut issuer, token rotation (predecessor-linked, fresh id)
- **token** token model (identity, task, capabilities, expiry)
- **validator** cross-boundary validation, `AttestationRejection` enum, audit ring buffer, checkpoint and restore of rotation state
- **launcher** fail-closed sandbox boot (enforced and bypassed modes)
- `tests/mik_5223_acs.rs` 11 real AC tests (AC1 through AC11)

Wired `pub mod attestation;` into `src/lib.rs` (alphabetical, first unconditional module after the feature-gated `a2a`).

## Bet
Implements **B1-IDENT** (signed per-action attestation tokens; unique attribution for every agent and tool action).

## Status
Builds clean against current `main`. All 20 unit tests plus 11 AC tests green; clippy `-D warnings` clean; fmt applied. Excludes branch junk (worker-cargo-shims, agent-stdout dumps, stale Cargo.lock churn).

**Needs review before wiring into the request path** the module is present and tested but not yet invoked by the gateway request and dispatch flow. Draft for review, do not merge.

### DIFF
--- a/src/attestation/launcher.rs
+++ b/src/attestation/launcher.rs
@@ -1,0 +1,357 @@
+//! Fail-closed sandbox boot gate with attestation injection (MIK-5223 AC.1,
+//! AC.6) — no token, no start.
+//!
+//! Both substrates — gVisor (`runsc`) on Ubuntu and Apple containerization on
+//! macOS — expose OCI hook lifecycle points; the launcher injects the token
+//! at the `createRuntime` hook through one shared code path, so the token
+//! flow is byte-identical regardless of substrate.  The only substrate
+//! difference is the runtime label stamped on the handle.
+//!
+//! Rollback: `SYMPHONY_PLUS_ATTESTATION=0` boots without a token (the
+//! sandbox stays isolated but loses identity attribution).  Any other value,
+//! including unset, enforces fail-closed.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use chrono::{DateTime, Utc};
+
+use super::token::TokenClaims;
+use super::validator::{AttestationRejection, AttestationValidator};
+
+/// Rollback feature flag: `0` disables boot-time attestation enforcement.
+pub const ATTESTATION_FLAG_ENV: &str = "SYMPHONY_PLUS_ATTESTATION";
+
+/// Environment variable injected into the sandbox at the OCI `createRuntime`
+/// hook carrying the encoded attestation token.
+pub const TOKEN_ENV_VAR: &str = "SYMPHONY_ATTESTATION_TOKEN";
+
+/// Boundary label used for boot-time validations in the audit ring buffer.
+pub const BOOT_BOUNDARY: &str = "sandbox_boot";
+
+/// Sandbox substrate (MIK-NEW.RUNTIME-A.6).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Substrate {
+    /// gVisor (`runsc`) on Ubuntu.
+    GvisorLinux,
+    /// Apple containerization framework on macOS.
+    AppleContainerization,
+}
+
+impl Substrate {
+    /// OCI runtime label for diagnostics; not part of the token flow.
+    #[must_use]
+    pub fn runtime_label(self) -> &'static str {
+        match self {
+            Self::GvisorLinux => "runsc",
+            Self::AppleContainerization => "apple-containerization",
+        }
+    }
+
+    /// OCI hook at which the token is injected — the same lifecycle point on
+    /// both substrates, keeping the flow identical.
+    #[must_use]
+    pub fn token_injection_hook(self) -> &'static str {
+        "createRuntime"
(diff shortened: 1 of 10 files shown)


MikkoParkkola:
## Adversarial review + minimal wiring (MIK-5223, B1-IDENT)

Reviewed the dormant attestation module against the standard attack classes, then wired it to a real action path behind a fail-safe seam. **DO NOT MERGE yet** — findings below are for operator decision; the wiring is opt-in and zero-cost when unset.

### Adversarial findings

**Resists (verified):**
- **Token forgery** — `BnautAttestationSigner::verify_bytes` uses `Mac::verify_slice` (constant-time); a forged token fails signature. `validator.rs:310`. [V]
- **`alg=none` / algorithm confusion** — the JWT-style `algorithm` claim is *never* trusted; validation always HMAC-SHA256s. The classic alg-swap attack does not apply. `validator.rs:300-322`. [V]
- **Parse-before-verify** — claims are parsed only from the *verified* payload bytes; a forged payload's claims are used as audit hints only, never for authz. `validator.rs:311-314`. [V]
- **Fail-closed boot** — missing token gives `BootDenial::MissingToken`; default enforces. `launcher.rs:74-79, 201-208`. [V]

**Gaps (operator should weigh before relying on this as an authz control):**
1. **[H] No capability enforcement at the boundary.** `validate_boundary_call` checks authenticity/expiry/rotation but never compares the token's `capabilities` allow-list against the requested action. A valid token minted for `["read"]` authorizes *any* tool. The allow-list is currently decorative. `validator.rs:258-351`. [V]
2. **[M] Replay within TTL.** No per-token uniqueness tracking on the validator — a captured, unexpired token replays indefinitely until `expires_at`. (The gateway has a *separate* nonce store on `gateway_invoke`, but attestation itself has none.) Bearer-token-inherent, but worth a short TTL plus the existing nonce layer. `validator.rs:337-349`. [V]
3. **[M] No clock-skew leeway and no not-before.** Expiry is a hard `now > expires_at` with zero tolerance; `issued_at` is never enforced, so a future-dated token validates. Skew between issuer and gateway clocks can cause spurious rejects/accepts. `validator.rs:327-336`. [V]
4. **[M] No signing-key rotation.** The validator holds one key; `key_id` is stamped but never used to select a key. There is no overlap window for HMAC key rotation (distinct from *token* rotation, which works). Key compromise means full reissue, no graceful roll. `validator.rs:208-246, 310`. [V]
5. **[L] Unbounded retiring map growth.** Rotated-out token IDs are inserted into `retiring` and never pruned after the grace window passes; a long-lived validator doing many rotations leaks memory. `validator.rs:359-372`. [V]
6. **[L] Caller-supplied now.** `validate_boundary_call(now)` trusts the caller's timestamp. Safe only if every caller passes a trusted clock — the wiring below does (`Utc::now()`), but the API permits a spoofed early `now` to bypass expiry. `validator.rs:258-263`. [V]

### What was wired

Wired into the real `gateway_invoke` dispatch path (`MetaMcp::invoke_tool` -> `check_attestation`, run after nonce/replay protection, before kill-switch/dispatch). `gateway/meta_mcp/invoke.rs`. [I]

- `MetaMcp::with_attestation(validator, mode)` — opt-in builder. **Unset = zero-cost no-op**; existing callers byte-identical.
- New `AttestationMode { Observe, Enforce }` (`attestation/validator.rs`) — replaces a bool (keeps the struct lint-clean) and makes rollout intent explicit:
  - **Observe** (safe default when wired): validates + audits every presented `attestation` token at the `gateway_invoke` boundary, never blocks, so enabling on live traffic cannot break unattested calls.
  - **Enforce**: fail-closed, JSON-RPC `-32002` on missing/invalid token.
- Uses the gateway's **trusted clock** (`Utc::now()`), closing finding #6 for this path.

### Tests (`gateway::meta_mcp::tests::attestation_wiring`)
- `no_validator_is_a_no_op_even_without_token`
- `observe_mode_passes_invalid_token_but_audits_it`
- `enforce_mode_rejects_missing_token_fail_closed`
- `enforce_mode_rejects_forged_token`
- `enforce_mode_admits_valid_token`
- `observe_mode_admits_valid_token_without_auditing`

Verification: `cargo test --lib attestation` gives 26 passed (20 module + 6 wiring); pre-push full suite **2873 passed, 0 failed**; `cargo clippy --lib` clean; `cargo fmt` applied. [V]

### Recommendation
Wiring is safe to merge as **dormant/observe-only** (no behavior change unless an operator calls `with_attestation`). Before flipping any path to **Enforce**, address finding #1 (capability enforcement) — without it, attestation authenticates *who* but not *what*, so it is an identity signal, not an authz gate.

PR #315, opened 2026-07-02
chore(release): v3.0.0 Trust Fabric (multi-user OAuth isolation landed)

MikkoParkkola:
Holds the v3.0.0 version tag per operator decision: a shared/multi-user ("team") gateway must not ship while it serves one stored OAuth login per backend to every caller.

**The defect (confirmed in code, this is not the identity-propagation feature):**
- OAuth tokens are stored keyed by `(backend, resource)`, not by user — `src/oauth/storage.rs:154` `storage_key`.
- The token is held on `HttpTransport.oauth_client` and attached to every request caller-agnostically — `src/transport/http/mod.rs`.
- So on a shared gateway, user A's request to a personal OAuth backend (Gmail, Superhuman) can be served with user B's login = cross-user data exposure. The classic `oauth:` backend path does not use v3.0.0 identity propagation and does not fail-closed.

**This PR (version hold only, no behavior change):**
- `Cargo.toml`/`Cargo.lock`: `3.0.0` → `3.0.0-dev` so the release can't be tagged by accident. No `v3.0.0` tag exists; the release workflow is tag-gated, so nothing has published.
- `CHANGELOG.md`: 3.0.0 marked UNRELEASED (held) with the reason.

**Blocker:** MIK-6742 (fail-closed guard as the minimum bar; per-user OAuth token store as the real fix). MIK-6704 reopened.

All merged Trust Fabric work stays on main and correct — this only walks back the version string.


MikkoParkkola:
## v3.0.0 shipped — release verified across all channels

Merged to `main` (`5ec385ad`) and tagged `v3.0.0`. Release workflow `28668200936` completed with all jobs green; each published artifact verified against its own registry:

| Channel | Version | Verification |
|---|---|---|
| crates.io | 3.0.0 | registry newest_version = 3.0.0 |
| npm (mikkoparkkola scope) | 3.0.0 | dist-tags.latest = 3.0.0 |
| GitHub release | v3.0.0 | published, 6 assets (5 platform binaries + SHA256SUMS) |
| Homebrew tap | 3.0.0 | tap formula pinned to 3.0.0 |

Pre-publish gates: fmt check clean, clippy all-features with warnings-as-errors zero warnings, full test suite green (3242 tests, 0 failures on push), and a locked publish dry-run succeeded in the verify job before the real publish. All 6 branch-protection required checks passed.

PR #324, opened 2026-07-04
fix(cli): serve degrades to defaults when --config path is missing

MikkoParkkola:
## Problem

Glama's build smoke-test starts the published container with `mcp-gateway serve --stdio --config /config.yaml` (its placeholder parameter), but the Glama-generated Dockerfile never writes that file. An explicit `--config <path>` that does not exist is a hard error in `Config::load` (`src/config/mod.rs:181`), so serve exited non-zero and Glama reported a spurious "build failed".

## Fix

`serve_config_path()` downgrades a non-existent explicit `--config` to the normal no-config resolution (fallback locations, env vars, then defaults) with a stderr warning — **for the stdio serve path only** (`run_stdio_server`).

Scoped to stdio deliberately: stdio is a local pipe with no network listener, so starting on defaults cannot expose an unauthenticated network surface. The HTTP `run_server` path stays **fail-loud** — its defaults (`auth.enabled=false`, `meta_mcp.enabled=true`) would be a fail-open regression if an intended `--config` went missing. `audit`/`validate`/`doctor` keep `Config::load`'s fail-loud behavior.

## Review

Adversarial GPT-5.5 second opinion: round 1 BLOCK (flagged the HTTP fail-open when the downgrade was applied to both paths); round 2 after scoping to stdio + reverting HTTP: **LGTM / SAFE TO MERGE**, verified at 5d998b2c.

## Tests

- `serve_config_path_none_stays_none`
- `serve_config_path_missing_downgrades_to_none` (the Glama case)
- `serve_config_path_existing_is_preserved`

## Gates

cargo test (3 new) pass · fmt clean · clippy --all-features -D warnings clean · full CI green.

## Note

Fixes the code defect. The Glama listing turns green only once (1) the fix ships in a release binary and (2) Glama's Dockerfile pin advances off v2.9.0 — Glama downloads a pinned release binary, it does not build from source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **stdio serve boots without a mounted config (Glama build).** An explicit
+  `--config <path>` that does not exist was a hard error, so
+  `mcp-gateway serve --stdio --config /config.yaml` exited non-zero when the
+  file was absent — the exact shape of Glama's container build smoke-test,
+  which passes that placeholder while the generated Dockerfile never writes the
+  file. The stdio serve path now downgrades a missing explicit `--config` to
+  the normal no-config resolution (fallback locations, env vars, then defaults)
+  with a stderr warning. Scoped to stdio only: the HTTP server path stays
+  fail-loud so a missing intended config can never silently start with
+  authentication disabled.
+
 ## [3.0.1] - 2026-07-03
 
 Security hardening fast-follow for the 3.0.0 end-user identity-propagation
--- a/src/main.rs
+++ b/src/main.rs
@@ -512,12 +512,43 @@ fn apply_cli_overrides_and_validate(config: &mut Config, cli: &Cli) -> mcp_gatew
 }
 
 /// Run the gateway in stdio mode (newline-delimited JSON-RPC on stdin/stdout).
+/// Resolve the effective `--config` path for the **stdio** serve command only.
+///
+/// An explicitly-supplied path that does not exist is downgraded to `None`
+/// (fall through to `Config::load`'s fallback-location search, then env vars,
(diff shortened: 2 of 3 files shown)


MikkoParkkola:
## DoD evidence — stdio serve missing-config fix

### Acceptance criteria

- **GLAMA.FIX.1** stdio `serve` with no `--config` and no file present starts instead of exiting non-zero. It logs a warning to stderr and falls back to fallback locations, env vars, then built-in defaults.
  Verified by `serve_config_path_none_stays_none` and `serve_config_path_missing_downgrades_to_none` (src/main_tests.rs). PASS.
- **GLAMA.FIX.2** HTTP `serve` stays fail-loud: a missing or unreadable `--config` still errors before binding a listener. `run_server` calls `Config::load(cli.config.as_deref())` directly, unchanged. No network fail-open.
  Verified by code inspection plus second-opinion review. PASS.
- **GLAMA.FIX.3** An existing `--config` path is passed through unchanged for both transports.
  Verified by `serve_config_path_existing_is_preserved` (src/main_tests.rs). PASS.

### Gate verdicts (CI head 672d2ee8)

- Tests: pass (full suite 3252 passed / 0 failed, pre-push hook plus CI Tests job)
- Clippy (pedantic, -D warnings): pass
- Format (fmt --check): pass
- CodeQL, Secrets scan, Dependency audit, Security gate: pass
- Kani, Public claims, Helm, K8s, Windows: pass
- No unsafe (deny unsafe_code) unchanged; no new task markers added

### Scope

Change is scoped to stdio transport only (local pipe, no network listener). HTTP transport behaviour is unchanged and remains fail-loud. Three files touched: src/main.rs, src/main_tests.rs, CHANGELOG.md.

### Second opinion

Independent GPT-5.5 review: round 1 flagged an HTTP fail-open regression and inaccurate wording; both fixed by scoping the degrade-to-defaults path to stdio and correcting the warning text. Round 2 verdict: safe to merge.

PR #353, opened 2026-07-12
docs(readme): extract v3.1 identity deep-dive into its own doc

MikkoParkkola:
## What

The README carried a long v3.1 identity section inline. This moves the deep-dive into `docs/whats-new-v3.1-identity.md` and leaves a short summary plus a link in the README.

## Also

Removed the Microsoft Graph mention. Token exchange (RFC 8693) ships, but no Microsoft Graph backend capability is wired, so naming it overstated what the release does. The Gmail example stays, since that path is real.

## Contents of the new doc

Sequence diagram, the three outbound identity strategies, the five safety invariants, the `POST /auth/token` endpoint, and 2.x upgrade notes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 **One gateway between your AI and every tool it needs, without flooding the context window.**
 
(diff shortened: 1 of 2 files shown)


MikkoParkkola:
Superseded by #357. The README restructure in #357 rewrites the same identity region, and its Multitenant identity section now carries the deep-dive: docs/whats-new-v3.1-identity.md was salvaged verbatim into #357 (commit 1e1265e5) with a link from that section, so no content is lost. Closing to avoid a competing merge into the same README region.

PR #354, opened 2026-07-12
ci(security): add CWE-532 secret-leak lint gate and wire into release gate

MikkoParkkola:
## What

Adds a CWE-532 secret-leak lint (`scripts/dev/cwe532-leak-lint.py`) and wires it into the CI and release gates. It flags `#[derive(Debug)]` structs and log macros (`info!` / `error!` / `println!`) that interpolate secret-named fields, which is the leak-into-logs class fixed in #323.

## Relation to the existing secrets-scan

The current `secrets-scan` job (trufflehog) scans git history for committed secret values. This lint targets a different class: secrets reaching logs at runtime through `Debug` or format strings. Complementary, not a duplicate.

## Design notes

- Baseline keyed by a stable `(path, kind, field)` identity rather than line number, so it survives unrelated edits.
- `// ci-allow-secret-*` escape markers for reviewed exceptions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -281,6 +281,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - name: Test the leak-lint parser (fail on parser regression)
+        run: python3 scripts/dev/test_cwe532_leak_lint.py
       - name: Run CWE-532 leak lint
         run: python3 scripts/dev/cwe532-leak-lint.py src crates
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -270,11 +270,30 @@ jobs:
         with:
           extra_args: --only-verified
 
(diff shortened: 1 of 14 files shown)


MikkoParkkola:
CI diagnosis: the new `Secret-leak lint (CWE-532)` job is failing on its own PR — 2 new findings not covered by `cwe532-baseline.txt`:

- `src/identity_propagation/mod.rs:69` — `BackendDescriptor` derives `Debug` over `token_exchange_scope`
- `src/identity_propagation/mod.rs:217` — `IdentityPropagationConfig` derives `Debug` over `token_exchange_scope`

Both fields are `Option<String>` holding an RFC 8693 `scope` request parameter (e.g. `"read write"`), not a credential value — plausible false positive of the naming heuristic (`token_exchange_scope` contains `token`). Leaving this for author judgment rather than guessing on security-sensitive code: either add `// ci-allow-secret-debug: RFC 8693 scope string, not a credential` on each field if that's correct, or redact if there's a reason `scope` could carry sensitive data in some configs.

Not merging — required checks are red and this needs an author call, not an agent guess.

Generated with Claude Code

PR #355, opened 2026-07-12
fix(idp-audit): fail-closed on identity-propagation audit write failure (MIK-6740 hardening)

MikkoParkkola:
## What

Makes identity-propagation audit writes **fail-closed on the mint path**. When the transparency-log write fails while minting a per-user credential, the mint is aborted with `PropagationError::AuditFailed` and an HTTP error response, so no minted credential ever reaches a caller without a durable audit record.

## Why

Main's `audit_identity_propagation` in `src/gateway/router/backend_handlers.rs` is best-effort, and its own comment marks fail-closed as "a possible future hardening, not required for MIK-6740." This is that follow-up.

## Scope discipline

Only the **mint** path becomes fail-closed. The non-mint resolution path stays best-effort, matching the static-credential fallback used elsewhere for a non-required backend — a minted credential is the only case where a missing audit record is a compliance problem.

## Tests

Adds a `mod audit_fail_closed` unit suite: logger-disabled no-op, mint-write success, and a fail-closed contract test that spawns an fsize-limited child to force a genuine transparency-log write failure and asserts the `AuditFailed` abort. Rebased onto current main with the test-module conflict resolved (main's MIK-6710 transport tests and this suite both retained).

### DIFF
--- a/src/gateway/meta_mcp/invoke.rs
+++ b/src/gateway/meta_mcp/invoke.rs
@@ -3132,6 +3132,16 @@ mod identity_propagation_enforcement_tests {
             output.status,
             String::from_utf8_lossy(&output.stderr)
         );
+        // The child must also EXIT cleanly: a child that prints the marker and
+        // then aborts (panic/abort after the observation) must not read as a
+        // pass. Mirrors the two sibling fail-closed subprocess tests.
+        assert!(
+            output.status.success(),
+            "child printed the marker but did not exit successfully \
+             (status={:?}, stderr={})",
+            output.status,
+            String::from_utf8_lossy(&output.stderr)
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


MikkoParkkola:
Do-not-merge: codex review flagged a fail-open regression in a fail-closed PR.

`audit_identity_propagation(None, ..)` returns `Ok(())` instead of surfacing the audit-write failure. That means when the audit sink can't be written, credential minting proceeds silently with no audit record — the opposite of what this PR's title (fail-closed on identity-propagation audit write failure, MIK-6740) promises. An attacker or an ops incident that knocks out the audit sink gets a free pass on unaudited credential minting.

Fix direction: `None` (or any audit-write failure) should propagate an error that blocks the mint, matching the fail-closed contract for every other path in this PR.

Leaving this open, not merging, per standing review policy on identity-propagation changes.

Generated with Claude Code

PR #356, opened 2026-07-12
fix(security): salvage SSE SSRF + transport hardening onto current main

MikkoParkkola:
## What

Salvages three SSE/transport security hardening fixes from an abandoned worktree branch and rebuilds them on current main. The branch was based on an older tree, so these were cherry-picked and one of them needed manual conflict resolution against main's restructured transport code.

## The three fixes

1. **SSE endpoint SSRF bypass via authority-relative URLs.** The SSE message endpoint is backend-controlled and arrives on the stream. The old code classified it by string prefix (`starts_with("http://")`) and passed absolute values through unchanged. WHATWG URL resolution treats `//host/x` and backslash shapes as authority-relative references, so a malicious backend could return `data: //169.254.169.254/latest/meta-data/...` and the gateway would POST the JSON-RPC request plus the per-user identity credential header to an attacker-chosen internal host. Every endpoint shape is now resolved against the base URL first, then the resolved origin is checked for same-origin equality before any credential is sent.

2. **64 KiB bound on the SSE handshake buffer.** Defence in depth against an unbounded read during the handshake.

3. **Removal of the dead Tailscale identity-header auth path.** 198 lines of unused code that widened the attack surface without a live caller.

## Conflict resolution note

The SSRF fix (commit 1) touched `resolve_message_url` and a neighboring helper that main had independently rewritten, so the 3-way merge could not auto-carry the `same_origin` helper or the new function body. Both are restored in the final commit. The pre-existing `resolve_message_url_absolute_http` test asserted the old cross-origin pass-through, so it is updated to assert the secure cross-origin rejection instead.

## Deferred

A fourth commit from the source branch (percent-encode positional path params, LOW severity) conflicts on the capability executor, which main split from a file into a directory and rewrote. It is left out of this PR to keep the security fixes clean. It can be ported separately if wanted.

## Verification

- `cargo test --lib transport::http` — 43 pass
- Full `cargo test --lib` via the pre-push hook — green (fmt, clippy `-D warnings`, and the full lib suite)

### DIFF
--- a/src/transport/http/mod.rs
+++ b/src/transport/http/mod.rs
@@ -33,6 +33,15 @@ use crate::protocol::{
 use crate::security::validate_url_not_ssrf;
 use crate::{Error, Result};
 
+/// Origin equality per WHATWG (scheme + host + effective port). Used to enforce
+/// that an SSE-advertised message endpoint is same-origin as the SSE stream
+/// before any per-user credential is sent to it (SSRF + credential-exfil guard).
+fn same_origin(a: &Url, b: &Url) -> bool {
+    a.scheme() == b.scheme()
+        && a.host_str() == b.host_str()
+        && a.port_or_known_default() == b.port_or_known_default()
+}
+
 /// Detect the session-expiry signature in a transport error (MIK-5982).
 ///
 /// Matches three observed shapes:
--- a/src/transport/http/mod.rs
+++ b/src/transport/http/mod.rs
@@ -611,21 +611,47 @@ impl HttpTransport {
         ))
     }
 
-    /// Resolve a potentially relative message URL against the SSE URL
+    /// Resolve a potentially relative message URL against the SSE URL.
+    ///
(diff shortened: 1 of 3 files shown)


MikkoParkkola:
Do-not-merge: codex review flagged three issues in this salvage PR.

1. Removes the public `TailscaleIdentity` type — breaking change for any downstream consumer constructing/matching on it directly.
2. The 64KiB SSE chunk cap false-rejects legitimate larger SSE chunks (valid protocol traffic gets dropped, not just malicious oversized ones).
3. Possible credential leak on cross-origin redirect in the SSE transport-hardening path.

Each of these needs its own fix before this is safe to merge: keep `TailscaleIdentity` public (or do a proper deprecation cycle), raise/parameterize the SSE cap to match real-world chunk sizes, and strip/re-validate auth headers on cross-origin redirect rather than forwarding them.

Leaving this open, not merging.

Generated with Claude Code

PR #395, opened 2026-07-26
chore(config): remove dead idle_timeout instead of pretending to implement it

MikkoParkkola:
`idle_timeout` was parsed, documented in shipped examples as "Hibernate after 5 min idle", and read by nothing but a `Debug` formatter. One operator had it set on **24 backends**; it did nothing on all 24.

This removes it rather than implementing it, because the replacement (`stop_when_idle_for`, next in the stack) has different and narrower semantics: it is valid only where the gateway owns the process. Keeping both would leave two settings in the config, one working and one inert, which is worse than either alone.

Retired keys are rejected at load with a message pointing at the replacement, so an operator who had it set finds out rather than silently losing behaviour they thought they had.

## Ordering

**This must land before the feature PR**, which is why it sits below it in the stack rather than above. The other order exposes both settings simultaneously.

## SemVer

Removing a public `BackendConfig` field is an API change. Released as **3.4.0 (minor)** rather than 4.0.0, verified rather than assumed: crates.io reports **0 reverse dependencies**, and the crate's consumer is its own binary. Rationale recorded in the CHANGELOG.

`cargo test --lib` 3453 passed / 0 failed; `cargo clippy --all-targets --all-features` clean.

Third of a four-branch stack. Targets #394.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,60 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   previous number of calls. `max_attempts: 0` clamps to a single attempt rather
   than none.
 
+### Removed
+
+- **`BackendConfig::idle_timeout` removed** (Rust API only; config files unaffected). The field was
+  parsed, documented in `examples/gateway-full.yaml` as "Hibernate after 5 min
+  idle", and read by exactly one thing in the codebase: a `Debug` formatter.
+  Nothing enforced it. Operators set it in good faith — one real config carried
+  `idle_timeout: 10m` on a stdio backend whose child then ran for over three days
+  and burned 10.4 CPU-hours.
+
+  Two attempts to implement it were reviewed and rejected. The blocker is not
+  difficulty: the name spans stdio child-process lifetime, per-user session TTL,
+  HTTP connection pooling, and remote scale-to-zero. Closing an HTTP client
+  transport does not scale down the service behind it, so no single
+  implementation can be correct for every transport.
+
+  **Impact on existing configs: none at load time.** Config parsing tolerates
+  extra keys, so files carrying `idle_timeout` keep loading unchanged. The
+  gateway now logs a warning naming the key so its inertness is visible rather
+  than silent. Note that any command which rewrites the config will drop the key
+  along with surrounding comments, since the writer re-serialises from memory —
+  delete it by hand first if you care about the comments.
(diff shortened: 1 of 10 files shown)


MikkoParkkola:
## ⛔ Blocked on a release-versioning decision — not on code

CI is green (25 checks) and the change itself reviewed clean. What blocks it is the version number.

### The objection

Removing `pub BackendConfig::idle_timeout` in a **minor** bump. Two rounds of adversarial review rejected it, and the second answered the specific question I put to it:

> Publishing the exclusion in the same release that first relies on it only establishes policy **prospectively**; it does not make this removal backward-compatible. Release this as 4.0.0, or preserve the field through the 3.x line and apply the new policy to later releases.

That is correct on the facts. `CHANGELOG.md` has always carried an unqualified *"this project adheres to Semantic Versioning"*, and 3.3.2 shipped `pub mod config` with a public `BackendConfig`. Anyone on a caret range (`^3.3`) would receive 3.4.0 and fail to compile.

### What I already tried, and why it isn't enough

I added a **Versioning and stability** section to the README stating that SemVer covers the product surface (CLI plus config format) and not the Rust library API. That was the reviewer's own third option, and it is worth keeping regardless — the scope genuinely was never written down anywhere citable.

But it only binds **future** releases. It cannot narrow the contract that 3.3.2 already published.

The "zero reverse dependencies" argument does not close the gap either: it shows nobody *has* broken, not that breaking is permitted.

### The options

1. **Release as 4.0.0**, keep the removal. Honest under the contract 3.3.2 published. The cost here is essentially the number alone — binary product, zero reverse dependencies. The new policy then governs 4.x onward, which is what it can legitimately do.
2. **Keep `idle_timeout` as a deprecated field through 3.x.** Nothing breaks, but it defeats this PR: the dead key stays in the config beside `stop_when_idle_for`, which is the two-settings-one-inert state this exists to prevent.
3. **Ship 3.4.0 anyway**, as a deliberate, recorded override.

My recommendation is (1). Awaiting a decision — I am not changing a published version number unilaterally.

Blocks #396, which is stacked on this.

MikkoParkkola:
Superseded by #396 and the v3.4.0 release. Current main already contains the retired idle_timeout detector and tests, the public versioning note, the examples cleanup, and the implemented stop_when_idle_for replacement. Rebasing this older standalone version bump would reintroduce obsolete release state.

PR #424, opened 2026-08-16
fix(warm-start): retry until a backend's tool cache is populated

MikkoParkkola:
## The bug

Two daemons launched in the same second (`ps -o lstart`, 2026-08-14 01:16:50): `mcp-gateway` and `hebb-serve`. The gateway warm-starts each configured backend **once**. hebb was still loading its embedding models, so that one attempt hit a closed port:

```
Warm-start failed backend=hebb    error=Transport error: error sending request for url (http://127.0.0.1:39400/mcp)
Warm-start failed backend=netdata error=Transport error: error sending request for url (http://localhost:19999/mcp)
Warm-start failed backend=context7 error=Backend timeout: Request timed out
```

One missed attempt was permanent. `backend_tools_for_discovery` returns cached tools by presence, and only fills an *empty* cache when `allow_empty_cache_fetch` is set — which `code_mode_backend_candidates` sets only for queries already prefixed `server:tool`. A plain semantic query never fills it.

Result: hebb was invisible to `gateway_search` for 1.5 days of uptime while `gateway_execute("hebb:recall")` returned real data the whole time.

## The change

| | |
|---|---|
| Retry until the tool cache is populated | Fast phase 2s→30s jittered for 180s, then every 60s |
| `ensure_started`, not `start` | `start_entry` has no already-running check, so retrying it would spawn duplicate transports |
| Backend re-resolved by name per attempt | A config reload replaces the instance; a captured handle revives a discarded one |
| Attempt ceiling derived from the backend's own `timeout` | A fixed ceiling silently pre-empts slow backends, and every attempt being cut means never discoverable |
| Abort-on-drop guard on the tasks | Dropping a `JoinHandle` detaches a task rather than stopping it |
| Retryable errors enumerated | The slow phase never ends, so a permanent failure must stop it |

## Known limitations, named in the code

- **Transport errors arrive flattened into a string.** `stdio.rs` maps a spawn failure to `Error::Transport("Failed to spawn: …")`, so a mistyped command path is indistinguishable from a port that is not listening yet, and both retry. Fixing it means preserving typed errors across the transport boundary — every backend call path, not warm-start. Filed rather than bodged; string-matching the message would break on the first wording change.
- **An empty tool list cannot be reconfirmed.** `get_or_fetch_shared` stores an empty result with a fresh timestamp, so retrying re-reads the cache without asking the backend. A bounded retry was implemented, measured to be a no-op, and removed.

## Tests

16 warm-start tests, six of them mutation-verified — disabling the retry, making dormancy permanent, emptying the `Drop` impl, deferring to the idle reaper forever, and accepting an empty list each fail exactly one test and no others.

`cargo clippy --all-targets -- -D warnings` clean · 3523 tests pass · `cargo fmt --check` clean.

## Review

Design reviewed over three rounds before any code existed, then the code over five. Findings that changed the approach: reconciling from the health loop could never work (it probes only `is_running() || is_circuit_tripped()`, so a backend that never started is skipped forever); treating dormancy as permanent recreated the original outage; and the first attempt ceiling was too tight.

Reviewed by Grok (SHIP) and GPT-5.x (all in-scope findings addressed; the transport-boundary refactor is filed as follow-up).

### DIFF
--- a/src/backend/ops.rs
+++ b/src/backend/ops.rs
@@ -401,6 +401,16 @@ impl Backend {
         BackendLifecycle::NotStarted
     }
 
+    /// The per-request timeout this backend was configured with.
+    ///
+    /// Exposed so callers that wrap a backend call in a ceiling of their own can
+    /// derive it from the operator's setting instead of hard-coding a number
+    /// that silently pre-empts any backend configured to take longer.
+    #[must_use]
+    pub fn request_timeout(&self) -> std::time::Duration {
+        self.config.timeout
+    }
+
     /// Get backend status.
     ///
     /// Reports the canonical Shared slot's circuit/health state (MIK-6735
--- a/src/gateway/server/mod.rs
+++ b/src/gateway/server/mod.rs
@@ -1231,11 +1231,19 @@ impl Gateway {
 
         // Warm-start backends: connect + prefetch tools into cache
         // If warm_start list is empty, warm ALL backends (makes list/search fast)
-        {
+        // Bound, not discarded: the returned guard aborts the retry tasks when it
+        // drops, so letting it fall out of scope here would cancel warm-start the
+        // instant it began.
+        let _warm_start_tasks = {
             let warm_start_list =
                 build_warm_start_list(&self.backends, &self.config.meta_mcp.warm_start, true);
-            spawn_warm_start_task(&self.backends, warm_start_list, WarmStartMode::Http);
-        }
+            spawn_warm_start_task(
+                &self.backends,
+                warm_start_list,
+                WarmStartMode::Http,
+                Some(&shutdown_tx),
+            )
+        };
 
         // Start health check task
         let backends_clone = Arc::clone(&self.backends);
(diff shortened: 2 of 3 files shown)


MikkoParkkola:
Follow-ups found during review of this change, recorded here because they are deliberately out of its scope. Each is verified at source, not speculative.

**1. Transport errors are flattened into a string.** The stdio transport maps every spawn failure to `Error::Transport(format!("Failed to spawn: {e}"))`, destroying the `io::ErrorKind`. A mistyped command path is therefore indistinguishable from a port that is not listening yet, and warm-start retries both. Fixing it means preserving typed errors across the transport boundary — every backend call path. The limitation is stated in `is_readiness_error`'s doc comment rather than hidden.

**2. An empty tool list cannot be reconfirmed.** `get_or_fetch_shared` stores an empty result with a fresh timestamp, so a retry re-reads the cache without contacting the backend. A bounded retry was implemented here, measured to be a no-op, and removed. Confirming an empty list needs a cache-bypassing refetch; `CachedMetadata` exposes no invalidation.

**3. The idle reaper handle has no abort-on-drop ownership.** Pre-existing, untouched by this change: `run_stdio` aborts it only on the EOF path, so cancelling `run_stdio` anywhere else detaches the task instead of stopping it. This change adopts a drop guard for its own tasks; the reaper could use the same treatment.

**4. The periodic health loop is absent from stdio mode.** It is spawned only in the HTTP path, and it probes only backends that are `is_running() || is_circuit_tripped()` — so a backend that never started is skipped forever, and disabling health checks removes it entirely. This is why the reconciliation in this change lives in warm-start rather than in that loop.

**5. `chains/retry.rs` documents full-jitter backoff and does not implement it.** It sleeps the exact exponential delay. Chain authors reading the doc will believe concurrent retries are spread out when they are synchronised. This is why warm-start does not reuse that helper.

PR #450, opened 2026-08-28
fix(security): close unauthenticated admin, provision a credential, and authorize at dispatch

MikkoParkkola:
Closes the externally reported origin-validation issue (CWE-346, reported by Pluto Security and confirmed at source), together with the three gaps that fixing it exposed. 133 commits since `v3.4.0`.

## Why these ship together

Removing admin from the anonymous identity leaves a default install with no admin path at all. Shipping that alone would replace one problem with another, so credential provisioning ships in the same release.

## What changes

**1. Close unauthenticated admin at `/mcp`** — `docs/design/origin-validation-anon-admin.md`

A caller with no credential can no longer reach the tool endpoint from a web page, nor hold admin. Adds Host/Origin validation and stops the disabled-auth identity from being granted admin and every backend.

**2. Provision an admin credential at install** — `docs/design/provision-admin-credential.md`

`init` generates an admin token, writes it owner-only, and says where it is. No YAML editing, no restart. `/health` and `/mcp` stay listed under `auth.public_paths`, so an already-configured MCP client keeps working; management does not.

**3. Unauthenticated network posture** — `docs/design/unauthenticated-network-posture.md`

Refuses to serve HTTP on a non-loopback bind with auth disabled, refuses a reload that would open the tool surface to the network, and creates credential files owner-only rather than relaxing permissions after the fact. The Windows file-permission gap is documented and warned about once.

**4. Authorize at the dispatch chokepoint** — `docs/design/authorize-at-dispatch.md`

Found during review of the above, not externally reported. `authorize_tool_target` was reached only from the router, and only for three call shapes; every other tool name authorized nothing. Playbook steps reached backends through a path where backend scope, per-client tool scope, global tool policy, mTLS policy and agent scope were never consulted. Authorization now runs at the single point all dispatch funnels through, ahead of the cache and the nonce check.

## Scope

Each design doc carries its own `FOR` / `OUT` statement and a threat model naming the adversaries it addresses and those it does not. Deliberately out of scope, tracked separately:

- authentication on by default for tool invocation — gating every path breaks already-configured clients
- the destructive-confirmation gate's fail-open path (MIK-7246)
- rate limiting for the anonymous identity
- capabilities that accept a caller-supplied URL (MIK-7247)
- readiness reported before the capability backend has loaded (MIK-7268)

## Evidence

- 38 `fix(security)` and 12 authorization fix/test commits; 23 design documents; ~20 test commits
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib` green
- `scripts/dev/first-run-smoke.sh` completes a routed capability call end to end and now asserts an upstream reading rather than a well-formed envelope — a routing failure previously returned a success envelope carrying `isError`, so the smoke reported green without reaching the capability
- design and test plans reviewed across two independent vendors before implementation


MikkoParkkola:
## Closure evidence

**Merged:** `5d25f104` (squash). Content verified on `main`: a tree diff from the branch tip to `origin/main` lists only files from PRs #426–#436, none from this change — every file this branch touched is byte-identical on the trunk.

### Acceptance criteria

| criterion | verdict | evidence |
|---|---|---|
| Admin surface rejects unauthenticated callers | PASS | dispatch-time authorization; covered by the router test suite |
| A credential is provisioned at startup, and its absence is fatal | PASS | BREAKING entry recorded in `CHANGELOG.md` under Unreleased |
| Sampling and elicitation reach only the session that asked | PASS | `send_to_session` targeted delivery; `e033da38`, contract documented in `685a2c0c` |
| A prompt that cannot be delivered releases its pending entry | PASS | `92fba6ef` |
| An undeliverable prompt returns `NoSession` rather than waiting out the timeout | PASS | `4a181b47` — the POST handler no longer holds its own receiver alive |
| A symlinked config reloads when its target is edited | PASS | `a97f1857`, `764c06c9` — the link target is resolved on every event |

### Gates

- **Tests:** `cargo test --all-features` — 4284 passed, 0 failed, 34 suites.
- **Static:** `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo fmt --check` clean.
- **CI:** all required checks green on `e4bc2f03` (Tests, build, Clippy pedantic, Windows check, Helm package/lint/render, secret-leak lint). Docker green.
- **Review:** dual-vendor adversarial review, round 10 on `e4bc2f03` — SHIP from two independent vendors (gpt, kimi), zero findings. Ratified against diff hash `fe0e3387…` before push.

### Residual

A symlink retargeted to a directory outside the watched set is not picked up, because the watcher cannot add directories from inside its own callback. Tracked in #453; referenced from the code comment and the changelog entry.

Four non-blocking improvements were raised in the final review round and deliberately left out of scope: separating session creation from stream subscription, a filesystem-backed watcher test for in-place target writes, factoring the duplicated register/send/cancel flow, and a startup-computed watch set for the event filter.

PR #460, opened 2026-08-28
docs(security): support only the latest release

MikkoParkkola:
Replaces the stale supported-versions table (2.6.x / 2.7.x) with the policy actually in force: only the latest release is supported, and security fixes ship forward rather than being backported.

Also names GitHub private vulnerability reporting as the preferred channel — both advisories published today arrived that way — with email kept as the fallback.

Scope: SECURITY.md only. No code, no workflow changes.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,23 +2,31 @@
 
 ## Supported Versions
 
-| Version | Supported          |
+Only the latest release is supported. Security fixes ship in a new release;
+they are not backported to earlier versions.
+
+| Version | Supported |
 | ------- | ------------------ |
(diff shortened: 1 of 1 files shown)


MikkoParkkola:
Review record: gpt-review (codex, adversarial) — SHIP: "the policy is clear, future-safe, consistently numbered, and contradicts no published advisory". Second-vendor review could not be obtained: kimi-review returned HTTP 429 (subscription rate limit) on three attempts across two runs, and grok is quota-exhausted. Recorded rather than worked around; the change is documentation-only and leaving a support table that names 2.6.x/2.7.x is itself misleading to anyone arriving from today's advisories.

PR #465, opened 2026-08-30
fix(meta_mcp): parallelize and bound prompts/list and resources/list backend fan-out

terafin:
## Summary

- `prompts/list` and `resources/list` aggregated every backend's prompts/resources by **sequentially awaiting each backend** (`handle_prompts_list` in `src/gateway/meta_mcp/protocol.rs`, `handle_resources_list` in `src/gateway/meta_mcp/resources.rs`).
- On a gateway with 50+ backends, a single slow or hung backend delayed the whole request by its full transport timeout. A client issues `prompts/list` + `resources/list` on every (re)connect; with one slow backend those blew past the client connect timeout, the client aborted, and the gateway's late response surfaced as `Received a response for an unknown message ID`, causing a reconnect loop.
- `tools/list` is unaffected because it is served from cached counts/surfaced tools and never fans out to backends synchronously.
- Fetch each backend's prompts/resources **in parallel** via `futures::future::join_all` (already used in `backend/registry.rs`), so total latency is bounded by the slowest backend rather than the sum of all backends.
- **Bound each backend fetch with `tokio::time::timeout`** (already used across the codebase) so a hung backend is skipped after 10s instead of blocking the whole request for its full transport timeout (often 120s). This is the piece that actually fixes the hang: a single dead backend no longer stalls every reconnect.
- Backends that fail or time out are logged and skipped, preserving existing behavior.

## Real-world validation

Deployed the patched binary against a live gateway with 56 backends. Measured `prompts/list` + `resources/list` through the meta-MCP endpoint:

| scenario | `prompts/list` | `resources/list` |
|---|---|---|
| upstream binary, serial fan-out | **120s+ timeout** | **120s+ timeout** |
| patched (parallel + timeout), one backend disabled | **~2.8s** | **~1.4s** |

The one disabled backend was hung (health probe timed out repeatedly). This is the exact failure the timeout addresses: without the timeout, that hung backend stalls every `prompts/list`/`resources/list` for its full 120s transport timeout; with it, the request completes in seconds.

## Validation

- `cargo check` clean
- `cargo clippy --all-features -- -D warnings` clean
- `cargo test --lib gateway::meta_mcp` — 390 passed
- `cargo test --lib gateway::meta_mcp::protocol` — 12 passed
- `cargo test --lib gateway::meta_mcp::resources` — 9 passed

## Declared gap

A backend whose fetch exceeds the 10s per-backend bound is skipped for that request, so its prompts/resources are absent from that response until it recovers. This is the correct trade-off: availability of the whole list over completeness. A gateway-level cache (mirroring the per-backend `CachedMetadata`) could avoid re-touching a slow backend entirely on warm reconnects; this PR does not add one, staying with the existing per-backend cache convention.

## First contribution

Thanks for a great project :) Finally found a great an aggregator that works for my self hosted stuff!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   surfaced as an "unknown message ID" error. Backends are now fetched in
   parallel and each fetch is bounded by a short timeout, so a slow or hung
   backend is skipped instead of blocking the list.
+- **The aggregation timeout is configurable** via
+  `meta_mcp.prompts_resources_fetch_timeout` (default `10s`). Operators with
+  unusually slow backends can raise it without a code change.
+- **A cancelled transport request no longer strands its `pending` entry.**
+  When an outer timeout drops an in-flight stdio or WebSocket request before
+  the transport's own request timeout fires, a RAII guard removes the entry
+  from the transport's `pending` map on drop, so a late response finds no
+  dangling sender and the map does not grow across reconnect loops.
 
 ## [3.5.0] - 2026-08-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **`gateway_search` returns L0 by default** (MIK-7084): tool name, one-line purpose, and score. `detail=l1` adds signature, when-to-use, and required params; `detail=l2` returns the full `input_schema`. `include_schema=true` still maps to L2 and is deprecated, not removed. Ranking diagnostics (`ranking` reasons and signals) are omitted unless `explain=true`. `gateway_search_tools` also omits `ranking` unless `explain=true`.
+### Fixed
+
+- **`prompts/list` and `resources/list` no longer stall on a slow or hung
+  backend.** Both handlers aggregated every backend sequentially, so a single
+  backend that was slow to answer held the whole request for its full transport
+  timeout (often 120s). On a gateway with 50+ backends this blew past client
+  connect timeouts on every (re)connect, and the gateway's late response
+  surfaced as an "unknown message ID" error. Backends are now fetched in
+  parallel and each fetch is bounded by a short timeout, so a slow or hung
+  backend is skipped instead of blocking the list.
 
 ## [3.5.0] - 2026-08-28
 
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -892,6 +892,16 @@ pub struct MetaMcpConfig {
     /// Tool cache TTL.
     #[serde(with = "humantime_serde")]
     pub cache_ttl: Duration,
+    /// Per-backend bound on how long `prompts/list` and `resources/list`
+    /// aggregation waits for a single backend before skipping it.
+    ///
+    /// The aggregate fetch runs all backends in parallel (see the meta-MCP
+    /// handlers), so this bounds the whole request by the slowest backend
+    /// rather than the sum of all backends. A backend that exceeds the bound —
+    /// slow OR hung — is skipped for that response; its prompts/resources
+    /// reappear once it recovers. Defaults to 10 seconds.
+    #[serde(default, with = "humantime_serde")]
+    pub prompts_resources_fetch_timeout: Duration,
     /// Backends to warm-start on gateway startup.
     #[serde(default)]
     pub warm_start: Vec<String>,
@@ -919,6 +929,7 @@ impl Default for MetaMcpConfig {
             enabled: true,
             cache_tools: true,
             cache_ttl: Duration::from_secs(300),
+            prompts_resources_fetch_timeout: Duration::from_secs(10),
             warm_start: Vec::new(),
             surfaced_tools: Vec::new(),
             projection_mode: crate::projection::ProjectionMode::default(),
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -109,6 +109,10 @@ fn default_routing_profile() -> String {
     "default".to_string()
 }
 
+fn default_prompts_resources_fetch_timeout() -> Duration {
+    Duration::from_secs(10)
+}
+
 #[derive(Default, Deserialize)]
 #[serde(default)]
 struct EnvFileConfig {
@@ -900,7 +904,10 @@ pub struct MetaMcpConfig {
     /// rather than the sum of all backends. A backend that exceeds the bound —
     /// slow OR hung — is skipped for that response; its prompts/resources
     /// reappear once it recovers. Defaults to 10 seconds.
-    #[serde(default, with = "humantime_serde")]
+    #[serde(
+        default = "default_prompts_resources_fetch_timeout",
+        with = "humantime_serde"
+    )]
     pub prompts_resources_fetch_timeout: Duration,
     /// Backends to warm-start on gateway startup.
     #[serde(default)]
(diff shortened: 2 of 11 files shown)


terafin:
Sorry for the ping. @MikkoParkkola - any chance you could review this? Happy to take any feedback!

MikkoParkkola:
@terafin thank you for the contribution! I'm happy you find the project useful. Apologies for the delay; I was a bit behind on my messages. I am reviewing it now, and you can expect comments shortly.

MikkoParkkola (changes requested):
Thanks again for this, @terafin. I've reviewed the implementation and like the direction. Parallel calls are the right approach here, and the numbers from your 56-backend setup were helpful.

A few changes before we merge:

- Please make the timeout configurable, keeping 10 seconds as the default. It applies to each backend's `prompts/list` and `resources/list` call, including startup and initialization, so some setups may need a higher limit.
- Please make the transport cleanup cancellation-safe. An outer timeout can currently leave a stdio or WebSocket request in the `pending` map.
- Please add a test with one fast and one hanging backend, confirming that the fast result is returned within a single timeout.
- Please run `cargo fmt --all` and Clippy; the new test code currently has a few issues.

This is a useful and focused improvement. Thank you for taking it on!

terafin:
@MikkoParkkola — thank you for the review! All four points are addressed in `8e1e2b9`:

1. **Configurable timeout.** The 10s per-backend bound is no longer hardcoded. Added `meta_mcp.prompts_resources_fetch_timeout` (humantime duration, default `10s`) and wired it through the `MetaMcp` builder in `build_meta_mcp`, so startup/init and every `prompts/list`/`resources/list` aggregation honor it. Operators with slow backends can raise it without a code change. I also collapsed the two per-list timeouts into one shared bound since they were always set together.

2. **Cancellation-safe transport cleanup.** An outer aggregation timeout can drop an in-flight stdio/WebSocket `request()` future before the transport's own request timeout fires, stranding its `oneshot::Sender` in the `pending` map for the transport's lifetime. Added a small RAII `PendingRequestGuard` (in `transport/mod.rs`) that removes the entry on `Drop`, and used it in both `StdioTransport::request` and `WebSocketTransport::request`. On success the reader task has already removed the entry (no-op); on internal timeout the guard removes it alongside the returned error. HTTP is unaffected (it has no `pending` map).

3. **Fast + hung test.** `prompts_list_fast_backend_not_stalled_by_hung_one` registers one responsive streamable-http backend and one that sleeps 60s, and asserts the fast backend's prompt is present within a single aggregation timeout. (Note: the bound must comfortably exceed cold-start — at 100ms the fast backend's own startup got skipped too. It uses a 1s bound.)

4. **fmt + clippy.** `cargo fmt --all` and `cargo clippy --all-features -- -D warnings` are both clean. Full meta_mcp suite: 402 passed. Transport suite: 134 passed (incl. the existing `request_cleans_pending_entry_when_write_fails`). Also dropped a dead `MockMcpBackend.name` field the test build warned on.

Happy to adjust anything further.

terafin:
Review follow-up: tests added for both behaviors, per your ask.

**1. Config field (deserialization + default)** — `src/config/tests.rs`:
- `meta_mcp_prompts_resources_fetch_timeout_deserializes_from_yaml` — `prompts_resources_fetch_timeout: 5s` parses to a 5s `Duration`.
- `meta_mcp_prompts_resources_fetch_timeout_omitted_in_yaml_keeps_default` — a present-but-partial `meta_mcp:` section (only `cache_ttl`) still gets the documented **10s** bound. This test caught a real bug: the nested `MetaMcpConfig` with `#[serde(default)]` + `humantime_serde` defaulted an absent duration to **0ns**, timing out every backend — fixed with `#[serde(default = "default_prompts_resources_fetch_timeout", ...)]`.
- `meta_mcp_prompts_resources_fetch_timeout_defaults_to_ten_seconds` — `Config::default()` yields 10s.

**2. Cancellation-safe transport cleanup** — `src/transport/stdio.rs`:
- `cancelled_request_does_not_strand_pending_entry` — spawns a real `prompts/list` request against a child that answers `initialize` but never `prompts/list`, then `abort()`s the task mid-await (the exact shape an outer aggregation timeout takes). Asserts `pending` is empty after the abort. **Verified it fails without the `PendingRequestGuard`** and passes with it.
- `pending_request_guard_removes_entry_on_drop` — isolated RAII unit test for the guard's `Drop`.

All green: `cargo fmt --check` clean, `cargo clippy --all-features -- -D warnings` clean, config/meta_mcp/transport suites pass.

terafin:
https://github.com/MikkoParkkola/mcp-gateway/pull/465#issuecomment-5531954335

MikkoParkkola (approved):
Follow-up commits 8e1e2b9 and e5534bb address the four review points: configurable `meta_mcp.prompts_resources_fetch_timeout` (default 10s, omitted YAML no longer deserializes to 0ns), `PendingRequestGuard` on stdio/websocket pending maps, mixed fast/hanging backend tests, and fmt/clippy.

Merging.

MikkoParkkola:
Merged — thank you, @terafin. All four requested changes are in, and I checked each against the code rather than the description:

- **Configurable timeout.** `meta_mcp.prompts_resources_fetch_timeout`, default 10s, applied per backend call, with three config tests covering deserialization and the default.
- **Cancellation-safe cleanup.** `PendingRequestGuard` with a `Drop` impl is the right shape — it replaces three hand-rolled `pending.remove` sites across stdio and WebSocket, and `write_message(...).await?` now leans on the guard rather than an explicit removal on the error path. `cancelled_request_does_not_strand_pending_entry` aborts a real in-flight task against a backend that never answers, with the internal timeout set far beyond the abort, so only the guard can do the cleanup. That is a genuine cancellation test rather than a `drop` of a pinned future.
- **Fast + hung test.** Present, asserting both that the fast backend's prompt comes back and that the hung one is skipped.
- **Formatting and Clippy.** Clean locally and green in CI.

I also checked the layer underneath, since a cancellation fix is only as good as what it wraps: `CachedMetadata`'s `FetchPermit` already clears `in_flight` and wakes waiters in its own `Drop`, so a cancelled leader does not poison the cache. The transport `pending` map was the one remaining leak, and this closes it.

**Two follow-ups on my side, in #474 — no action needed from you.**

An independent review of the diff caught that the three timing assertions bound at 10 seconds while the aggregation timeouts they exercise are 100ms and 1s. The backends in those tests carry their own 5-second timeout, so the pre-parallel serial path would also have finished inside 10 seconds — the assertions could not fail for the reason they exist. #474 moves them to 2s and 3s, below that serial bound and with room for slow CI, and each failure message names the bound it discriminates. No production code changed.

The same PR credits you: the three changelog entries now name you, and `AUTHORSHIP.md` — which recorded that the project had no contributors besides the maintainer — gains a `Contributors` section. That stops being true with this PR, which is a good problem to have.

Two smaller suggestions from the same review, recorded rather than requested — neither is being asked of you:

- a WebSocket counterpart to the stdio cancellation test (the guard is shared, so the stdio test already exercises its `Drop`; the WebSocket call site is two lines of identical application)
- a fan-out case with two hanging backends, asserting completion within a single aggregation timeout

**One administrative loose end, and it is on me for merging before checking.** This repository is dual-licensed, so accepting outside code needs an explicit licence grant on record. `CONTRIBUTING.md` asks first-time contributors for a one-line CLA statement in the PR description — I could not find one here or in the commit trailers, and I merged anyway. Could you add this as a comment on this thread when you get a moment:

```
I have read and agree to the mcp-gateway Contributor License Agreement (CLA.md).
Signed-off-by: Your Name <your@email>
```

It is a one-time thing and covers anything you send in future. The terms are in [`CLA.md`](https://github.com/MikkoParkkola/mcp-gateway/blob/main/CLA.md) — it is a licence grant, explicitly not an assignment of copyright, so you keep ownership of your work. Until it lands, the contributors table records the CLA as outstanding rather than implying one.

Three quick rounds on a real fix from a 56-backend deployment — genuinely appreciated.

Issue #90, opened 2026-03-04
[FEATURE] [ROI:25x] CLI Bridge — Expose Gateway Tools as Composable CLI Commands


MikkoParkkola:
## CLI Bridge — Composable CLI Interface for Gateway Tools

**Source**: Eric Holmes "MCP is Dead, Long Live the CLI" blog post (Feb 2026)
**Insight**: CLIs compose (pipe, jq, grep), are human-debuggable, and require no background processes. MCP tools are isolated black boxes. The strongest argument against MCP is composability — but this is solvable by exposing MCP tools as CLI commands while keeping structured schemas.

### The Argument (and why we win both sides)

Holmes argues CLIs are superior because:
1. Composability: `terraform show -json | jq '.resource_changes[]'` — impossible with MCP
2. Debuggability: Run the same command the agent ran
3. No moving parts: Binaries on disk, not background processes

**What Holmes misses** (and what mcp-gateway already solves):
- Structured schemas prevent ambiguous CLI output parsing
- Tool discovery across 180+ tools without training data luck
- Centralized auth (one gateway, not N separate logins)
- **Webhooks and external triggers** — CLIs are request-response only. mcp-gateway supports event-driven execution via webhooks and external triggers. Try doing `gh webhook listen | trigger-agent-fleet` with pure CLI. This is fundamentally impossible without a persistent process.
- **Cross-environment**: MCP works in browsers, mobile, sandboxed — CLIs need shell access

### What mcp-gateway Should Build

A CLI bridge that exposes every gateway tool as a composable command:

```bash
# Direct invocation with structured output
mcp-gw exec fulcrum:linear_search_issues --query "MIK-556" | jq '.nodes[].title'

# Piping between tools
mcp-gw exec fulcrum:linear_search_issues --query "P0" | \
  jq '.nodes[].identifier' | \
  xargs -I{} mcp-gw exec fulcrum:linear_get_issue --id {}

# Human-readable format
mcp-gw exec fulcrum:linear_search_issues --query "research" --format table

# Tool discovery
mcp-gw list | grep linear
mcp-gw describe fulcrum:linear_create_issue

# Same auth, no re-login
mcp-gw auth status
```

### Architecture

1. **CLI Wrapper**: Thin CLI that connects to running gateway process (or starts one)
2. **Structured JSON Output** (default): Pipe-friendly, jq-compatible
3. **Human-Readable Format** (--format table/text): For debugging
[...]

PR #473, opened 2026-09-01
feat(protocol): v4.0.0 multi-round tool result readiness

MikkoParkkola:
An enforced content-integrity transform (Strip or Summarize) rebuilt the delivered result and dropped the fields that mark a multi-round exchange as still open. A caller received what looked like a completed call, and the continuation was lost.

## What changed

- The rebuilt envelope carries the round discriminator and, when the payload really is an input-required round, its handle. The handle is gated on the parsed protocol type rather than on a field name.
- Round descriptions cross verbatim. An unrecognized round type, a non-string discriminator and a malformed error flag are no longer filtered, coerced or defaulted, because every such rule rewrites a malformed round into a finished success.
- An absent error flag stays absent instead of defaulting to false.

## Scope

FOR: the enforced transform must not present an unfinished round as a completed call.
OUT: wiring the sealed continuation envelope, and the policy decision on withheld structured questions. Both are disclosed as open decisions in the v4.0.0 release-note draft.

## Evidence

- 3808 library tests pass; three new regression tests cover the handle, the invented-handle case and an unrecognized round type.
- Falsifier probes against the pre-fix and the superseded intermediate implementations fail on the expected assertions, and pass again after restore.
- cargo fmt --check clean; cargo clippy --all-targets -D warnings clean.
- Reviewed by GPT-5.x and by Grok; findings on the malformed-description handling and the cited commit trail are addressed in this branch.

---

## Scope amendment

The branch has grown well past the transform fix described above, and the title and Scope section
had not moved with it. Recording the change rather than leaving the description wrong.

FOR, as it now stands: v4.0.0 protocol readiness. The original transform fix, plus the multi-round
tool result (MRTR) work it exposed — the continuation handle and its state, the legacy-client
bridge, reclamation of abandoned in-flight exchanges, the continuation telemetry contract — and the
release documentation and gates that describe them.

OUT, unchanged: capacity constants, stateless continuations, and any per-principal quota on the
in-flight table.

Why it moved: the transform fix could not be stated correctly without saying what an unfinished
round is, and that definition had no owner in the tree. Each subsequent design settles one question
the previous one could not answer on its own. The alternative was a fix whose central term was
undefined.

## Design documents in this branch

| document | settles |
|---|---|
| `docs/design/2026-09-01-mrtr7-legacy-client-bridge.md` | how a client that predates the round protocol is prompted on a backend's behalf |
| `docs/design/2026-09-01-nfr-perf3-reclamation.md` | how abandoned in-flight exchanges are reclaimed, and how a soak observes it |
| `docs/design/2026-09-01-continuation-telemetry.md` | what the continuation counters mean and what bounds each series |

## Review status

Design review is dual-vendor per the delivery process. The reclamation design has a recorded
SHIP-WITH-FIXES from the GPT leg; its blocking finding — that pressure-triggered cleanup alone
does not bound how long expired state lives — is repaired in `b8bd5c21`, which splits the
bounded-count and bounded-lifetime obligations into two mechanisms and gives the reaped-expiry
counter a producer. The second-vendor legs are running; their verdicts will be recorded here before
this merges.

## Design decision made during implementation — protocol-era probe (`bb3727a2`)

Named here because it was not settled by any design document, and a reader who finds only the
commit would have no way to tell that.

`src/protocol/era.rs` has carried a complete, unit-tested `EraCache` since `390ab101` with no
production caller, and the acceptance tests at `tests/mik_7217_era_probe_acs.rs` were committed
ahead of the API it needed (`6f2d818e`, "held for the API"). The shape of `Backend::cached_era` is
therefore taken from those tests, because nothing else specifies it. That is a design decision taken
while writing the code, not a design that was reviewed, and the tests are its only specification.

What that means for a reviewer: the probe's placement, its 2s bound, and the choice never to cache
silence are all defensible and all unexamined by anyone but the author. They are worth reading as
proposals rather than as settled positions.

### Two limits on the cache's lifetime, recorded

Both are the same shape — the cache outlives the thing it describes — and neither blocks this
change. They are recorded rather than filed, because acting on either is a decision for whoever
writes the era design, and a ticket would only ask that question in a more expensive place.

- The cache is per-`Backend` and shared across every `PoolKey` slot, so under
  `session_mode = per_user` only the first slot probes and later slots inherit its verdict. Correct
  for stdio, where there is one peer. Incorrect for HTTP behind a load balancer with mixed-version
  upstreams, where the slots are not the same peer.
- `force_restart` does not clear the cache, so a restarted backend keeps the verdict taken from the
  process it replaced.

### Measurement

Full suite green at `bb3727a2`: 4821 passed, 0 failed, clippy clean, `cargo fmt --check` clean. The
2s probe bound is a per-start cost against a silent peer; the acceptance binary runs in 5.59s.

The tests were retrofitted against an implementation that did not exist when they were written, so
the free failing-first run was available and was taken: disabling the single `resolve_era` call
produced `0 passed; 6 failed`, each on its intended assertion, and the differential case failed on
its positive arm (`left: 0, right: 2`) rather than going vacuous. Restoring the call returned
`6 passed`, verified by re-running the tests.


MikkoParkkola:
## DoD evidence

**Scope (§P0).** FOR: an unfinished round stays unfinished through an enforced content-integrity transform. OUT: wiring the sealed continuation envelope; the `inputRequests` policy question. Both disclosed in the release notes as MRTR.3 and as an open operator decision.

**Reviews (§12, dual vendor).** Round 1: Codex `SHIP-WITH-FIXES` (2 findings), Claude Opus 5 `SHIP` (4 improvements). Confirmation pass: Codex `SHIP-WITH-FIXES` (1 HIGH), Claude Opus 5 `SHIP`.

| finding | disposal |
|---|---|
| Round description was rewritten rather than echoed | eliminated — `0892eea5` |
| Commit trail cited a superseded intermediate | corrected — `72bffd29` |
| Untyped `resultType`/`isError` cloned verbatim past the transform (HIGH) | eliminated — `93b56b9d`, the round is refused |
| Summary row claimed sealing that has no production caller | corrected — `779882e7` |
| Raw backend `requestState` bypasses the sealed envelope | out of scope, pre-existing whole-path gap — disclosed as MRTR.3 |

Two improvements declined as cosmetic: a shared fixture helper, and a state-only `input_required` case. The handle path does not read `inputRequests`, so a state-only round is already carried by the same branch the existing tests exercise.

**Gates.** `cargo fmt --check` clean · `cargo clippy --all-targets -- -D warnings` clean · `cargo test --lib meta_mcp` 417 passed, 0 failed · 5 new tests, each written before its repair and each observed failing first.

MikkoParkkola:
## DoD verdicts and open findings at 47ac3885

### Gates

| gate | verdict | evidence |
|---|---|---|
| §3 static | PASS | `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt` applied |
| §4 testing | PASS | every suite green — 3872 lib + 81 MIK-7212 acceptance, 0 failed |
| §5 change safety | PASS | no coverage drop; both repairs carry a test that fails without them |
| §6 regression | PASS | each repair's test was run against the pre-fix source and failed on the expected assertion |
| §7 documentation | PASS | DE-7 added to `docs/design/2026-08-30-mrtr-wiring.md`; criteria ledger updated |
| §12 peer review | **PARTIAL** | GPT leg returned SHIP-WITH-FIXES; the second vendor leg has not completed against this head |
| §11 stop-the-line | **BLOCKED** | four findings below are open, one HIGH |

### Acceptance criteria touched this round

- **MRTR.2 — MET.** An earlier revision of the row recorded UNWIRED on the ground that `Keyring::mint` had no production call sites. That was true when written and is not true now: mint runs at `src/gateway/meta_mcp/invoke.rs:388`. The correction came from review, not from noticing it here.
- **MRTR.9 — MET for the refusal, with an open defect in its payload.** The refusal fires; its `requiredCapabilities` data was being discarded at the response boundary and now survives, one key wide. The payload's *shape* is still contested — see finding 2.
- **MRTR.10a — UNWIRED, blocking.** The retry path is refused before a caller context is built, so the key discriminator is exercised only by hand-built fixtures.

Ledger now reads 95 criteria, 99 rows, 65 met or non-blocking, **34 blocking**. Cluster A remains the bulk of the remainder.

### Open findings — none repaired

1. **HIGH** `src/protocol/mrtr.rs:260` — the capability gate keeps only top-level capability names, so a URL-mode elicitation can be relayed to a form-only client. Mode-specific substructure is never inspected.
2. **MEDIUM** `src/gateway/meta_mcp/invoke.rs:442` — `requiredCapabilities` is built as an array where the spec calls for a `ClientCapabilities`-shaped object, and `-32021` is not mapped to HTTP 400. The discard half of this finding is fixed; the shape and status half is not.
3. **MEDIUM** `src/protocol/mrtr.rs:228` — a non-object `inputRequests` is treated as absent rather than malformed. Narrowed by this round's guard when no `requestState` is present, still reachable with one.
4. **MEDIUM** `src/gateway/meta_mcp/invoke.rs:444` — an unrecognised input method is reported as `-32021`, which tells a client it failed to declare a capability when the actual fault is upstream.

These are recorded rather than fixed because each changes what a refusal means to a client, which is a design question and not a patch. This PR is not release-ready.

MikkoParkkola:
## Review round closed: three findings repaired, one split out as blocking, one closed at source

Two commits since the last update, plus a ledger change.

**`a7943825` — a malformed `inputRequests` is now told from an absent one.** A present-but-not-an-object
`inputRequests` was read as an empty map, so a backend sending `"inputRequests": "oops"` produced a
valid state-only interim result instead of a refusal. Absence still means what it meant: a result
carrying only `requestState` is a legitimate turn where the backend wants to continue without asking
the user anything. Both readings are asserted in `ac_mrtr_7_a_completed_result_is_not_mistaken_for_one`.
Falsifier run against the pre-fix source: fails on its own assertion, passes after the repair.

**`MIK-7212.MRTR.9a` — new blocking criterion, split out of MRTR.9.** The capability gate refuses by
capability *name* and cannot express a *mode*. The specification states the requirement as a MUST —
"Servers MUST NOT send elicitation requests with modes that are not supported by the client" — and
fixes an empty `elicitation: {}` as equivalent to `{ "form": {} }`
([spec](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation#capabilities)).
`RequestShape::declared_capabilities` flattens the declaration to its names at the parse boundary, so
a URL-mode question relayed to a form-only client passes the gate by construction; nothing under
`src/` models elicitation modes at all. MRTR.9 stays closed — its per-type refusal is implemented and
tested, and no repair to it addresses modes. The two halves fail differently, which is why this is a
separate row rather than a reopening. Rationale in `docs/design/2026-08-30-mrtr-wiring.md` DE-8.

**One finding closed at source, unrepaired.** The claim that the refusal payload must be a
`ClientCapabilities`-shaped object carrying HTTP 400 is not supported by the page cited for it: the
MRTR specification contains neither the field name nor a status mandate, and says only that protocol
errors "SHOULD return a JSON-RPC error response with an appropriate error code and message". Recorded
in DE-9 rather than silently dropped.

**One finding deferred with its reason.** An unrecognised input method is refused with `-32021`, the
code meaning a capability is missing, when nothing is missing. That holds against the same sentence,
but changing a refusal's error code is an observable contract change on the path MRTR.9a rewrites;
it travels with that row rather than being done twice.

Coverage after the split: 95 criteria, 100 rows, 65 met or non-blocking, 35 blocking.

Validation on the current head: `cargo clippy --all-targets -- -D warnings` clean, every suite green.

Review status is stated rather than assumed: the GPT leg returned SHIP-WITH-FIXES and its findings are
disposed above. The second leg is incomplete — the Grok reviewer is returning `402 Payment Required,
usage balance exhausted`, so no verdict row exists for it, and the substitute leg is still running. The
dual-vendor gate is open until a second verdict is recorded.

MikkoParkkola:
## Second review leg: complete, four findings, two repaired in `5dd9ac8d`

The second vendor leg finished (process exit 0). Run output:
`~/.claude/data/reviews/runs/synthetic-20260902T075419Z-90936.md`, verdict line `SHIP`.

**Findings and their disposal**

| # | finding | disposal |
|---|---|---|
| 1 | The idempotency commit is gated on `from_result` classifying the result as interim, not on the backend having finished — so a result that claims `input_required` but is declined by the classifier is settled with the final-shaped *side effect executed* placeholder | **repaired** in `5dd9ac8d` |
| 2 | The response-cache write is gated on "a continuation was minted", which has the same hole one layer down | **repaired** with it — same predicate, and the intermediate boolean is gone |
| 3 | `InputRequired::undeclared` returns only the first offending entry, so a result naming several undeclared types is refused one at a time | recorded as an observation; per-entry checking is the intended behaviour and the refusal is terminal, so the cost is round-trips rather than correctness |
| 4 | The `MIK-7212.MRTR.10a` row claimed one function makes it impossible to move one derivation without the other — true of what the discriminator contains, not of where it is appended | **ledger corrected**; the row now says what the single function does and does not buy |

The repair adds `InputRequired::claims_input_required`, which reads the `resultType`
discriminator and nothing else. `from_result` answers a narrower question — can the gateway carry
this exchange — and declines shapes that do claim `input_required`. Both post-dispatch gates now
ask the backend's claim, because keying on the classification exempts exactly the shapes the
classification rejects. Design event recorded as DE-10 in
`docs/design/2026-08-30-mrtr-wiring.md`.

Test: `ac_mrtr_7_a_declined_shape_is_still_a_backend_that_stopped_to_ask`. Falsifier probe run —
reimplementing the predicate as `from_result(..).is_some()` fails it on the expected assertion,
and the test passes again once the source is restored. Full suite green, clippy clean.

**Review-gate status, stated rather than assumed.** Neither leg wrote a row to
`~/.claude/data/reviews/paired-ledger.jsonl`, so the ratification gate reads `MISSING` for both
regardless of the verdicts in the run files above. The Grok leg is separately unavailable
(`402 Payment Required`, usage balance exhausted). The verdicts here are evidence about the code;
they are not a satisfied gate, and this comment should not be read as one.

Ledger unchanged in count: 95 criteria, 100 rows, 65 met or non-blocking, 35 blocking.

MikkoParkkola:
## Cluster A: test plan committed (`251a7d9e`)

`docs/design/2026-09-02-mrtr-test-plan.md` — one row per `MIK-7212.MRTR.*` criterion: the case that
proves it, its V-model level, its type, and what makes that case able to fail.

**What the matrix found.** MRTR.9a has no case, and that empty cell is the finding rather than a
gap to be filled later: the criterion is ABSENT, not untested — the declaration type carries no
mode field, so a case written today would assert against a value that does not exist and pass
vacuously. It stays with DE-9 until the field is added.

Two rows (MRTR.5c atomicity, MRTR.6 replica routing) are the cluster's only two-replica fixtures;
the other nine run in-process. Writing the in-process rows first is not a scheduling preference —
MRTR.4 and MRTR.5a/5b exercise the same mint-and-verify surface the integration rows depend on, so
a defect there is cheaper to find before two processes are involved.

MRTR.6's "or fails explicitly" leaves the refusal's shape undecided — a named JSON-RPC code, or a
refusal reusing the MRTR.3 reason set. Recorded as an open design question rather than assumed,
with the case written to accept either as long as the failure is named.

Four rows (MRTR.2, .9, .10a, .10b) are marked as already held by existing `ac_mrtr_*` tests. The
review is asked to check that claim, not take it.

### Review gate: still OPEN, and for a second reason now

- `grok-review`: `API error (status 402 Payment Required): Grok Build usage balance exhausted`.
- `gpt-review`: could not be launched this session. A policy guard blocked the invocation with
  `SECURITY BLOCK: Command targets the EIS daemon socket (/tmp/eis.sock)` — nothing in the command
  referenced that socket — and then latched, classifying every subsequent long command as
  circumvention. Not worked around. Recorded as a guard misfire needing a look.

No verdict exists for this plan and none is claimed. Per §PA a verdict is a ledger row, and there
is no row.

MikkoParkkola:
## `NFR.COMPAT.3` waived on the record — `exposed_meta_tools` enforces in 4.0 (`158a9e00`)

Asked whether to warn in 4.0 and enforce in 4.1, enforce now and accept the break, hide enforcement
behind a second opt-in flag, or drop the field. The answer was **enforce now**.

The ledger row is now `N/A` **and says why in those words**: the requirement was set aside by the
person entitled to set it aside, not met. Recording it as a bare status change would have hidden a
real break behind a status code — an operator who edits nothing loses tools that used to answer.

Blocking count: **35 → 34**. The counter agrees (`scripts/release/count-release-criteria.py`, exit 0).

### What the waiver buys, and what it owes

The obligation transferred to the release was "tell an operator which tools they stop seeing". A
warning that a field is now enforced is useless without the list it restricts, so the breaking-change
entry now carries it — and reading the enforcement code to write that paragraph turned up three
things the earlier prose did not say:

- **Backend tools are unaffected.** `is_exposed` always exposes names outside the governed roster;
  surfaced backend tools are not meta-tools and the allow-list has nothing to say about them. The
  loss is confined to the 19 governed names.
- **`gateway_invoke` is not force-included.** An allow-list omitting it logs a warning and is then
  honoured, leaving a gateway that can list backends and invoke nothing. The warning is the only
  safety net, which is worth an operator knowing before they upgrade.
- **A typo drops one tool, it does not fail startup.** Unrecognised names are warned about and
  discarded, matching `surfaced_tools`.

Code Mode's `gateway_execute` sits inside the governed set rather than outside it, which is what
closes the escape hatch where that one tool stayed callable whatever the operator named.

### Cluster F: two decisions still open

Both are operator calls, not engineering, and both are among the cheapest blocking rows on the board:

- `NFR.COMPAT.4` — is a dual-role compatibility matrix required for 4.0.0? No dual-role harness
  exists; this is a meta-criterion over every other row and no row records both roles.
- `NFR.PERF.4` — the requirement says the Meta-MCP surface stays 14-16 tools, and
  `benchmarks/public_claims.json` records a documented 17-tool scenario. Nothing clamps the count,
  so the criterion is violated by the shipped configuration. Either the ceiling moves or the
  scenario does.

MikkoParkkola:
## Cluster A test plan reviewed — SHIP-WITH-FIXES, incorporated (`ed4e3dd8`, `ea0f0c0b`)

**Verdict, stated honestly: one vendor, not two.** `gpt-review` returned SHIP-WITH-FIXES with four
findings. `grok-review` returned `API error (status 402 Payment Required): Grok Build usage balance
exhausted` — its wrapper exited 0 with the error in the body, so per §PA that leg is an error, not a
verdict, and nothing was scraped from its text. The dual gate stays open on the second leg.

### Self-QA caught two of my own claims first

The plan submitted four rows as "already held by existing `ac_mrtr_*` tests" and asked the review to
check that claim. Checking it myself first broke two, in opposite directions:

- **MRTR.2 was overstated.** Both `ac_mrtr_2_*` tests build a `Keyring` by hand and assert on the
  minted token — confidentiality at unit level against a value the test constructed. The criterion is
  about what the gateway does with a *backend-supplied* state, and no test drives that path. A
  passthrough regression would leave every one of those tests green. The row is now a case still to
  write. It matters beyond the plan: MRTR.2 is `MET` in the ledger on code wiring, and the test
  evidence behind that citation is thinner than it reads.
- **MRTR.10b was understated.** It is held better than claimed, by `tests/mik_7216_mrtr_10_acs.rs` —
  both caches, the refusal, the control that final results still cache, malformed types. The plan
  cited the wrong file.

### What the review changed

**The MRTR.9a exemption did not survive, and it was the plan's headline finding.** I argued no test
could exist because the declaration type carries no mode field. That confuses the Rust type with the
wire: a client can declare form mode in raw JSON and the gateway can be driven to relay a URL-mode
elicitation *today*, with no new field anywhere. The case fails now — the free failure the criterion
was being excused from. The excuse was more comfortable than the test, which is what made it worth
having someone else read.

**The MRTR.5 fixture tested a mechanism we do not deploy.** I assumed two replicas over a shared
store. The accepted design says the opposite in terms — *key material is per process and is never
shared* (`docs/design/2026-08-30-shared-continuation-state.md:107-120`) — and cross-replica
single-use is satisfied by refusal, because a second process cannot open the envelope at all and
never reaches a ledger. Left as drafted, the test would have passed against a design nobody agreed to
build. MRTR.6 carried the same assumption and is corrected the same way.

**Two rows were single cases where the criterion is plural.** MRTR.1 tested both-fields-present when
a retry may legitimately carry one; MRTR.9 tested `sampling` when the criterion covers every
undeclared type. Both are now tables — neither single case could fail on the shapes it omitted.

One remark is recorded without action: the verdict line refers to an invalid oracle in MRTR.3, but no
finding block was filed for it and the row's four presentations each construct a value a correct
verifier must reject. Unverified at source, so neither fixed nor dismissed — it goes to the second leg.

### Where that leaves cluster A

Ten cases to write, not nine (MRTR.2 rejoins). Two need a second process; the rest run in-process,
and the in-process rows go first because MRTR.4 and MRTR.5a/b exercise the mint-and-verify surface
the two-process rows depend on.

MikkoParkkola:
## Status update — 4.0.0 release-criteria work on this branch

### Landed since the last update

| commit | what |
|---|---|
| `6daf020f` | server capabilities declare an `extensions` map, populated from implemented-and-enabled handlers |
| `73e3f27d` | locator record for the stdio negotiated-revision work, whose content is in `859f3094` |
| `3ddf2bde` | readiness board names which half of the `NFR.PERF.1` comparison is missing |
| `b9b8d804` | GH475 test plan repaired against the tree it describes |

### Deliberately red

`mik_7212_wire_11_cancelled_sampling_does_not_strand_pending_entry` (`src/gateway/proxy.rs:991`)
is a failing test written before its implementation. It is expected to fail until the
cancellation wiring lands; the transition to green is the evidence that the wiring works.
It is not a regression and should not be repaired, deleted, or routed around.

### Outstanding on this branch

- Response-cache keys hardcode `protocol_revision: None` at both construction sites, so entries
  are shared across protocol revisions. The fix is to thread the declared version through from
  the router; a second option that disables the cache wholesale was rejected as a product
  decision rather than a correctness fix.
- Tasks extension: implementation gated on a citable spec artifact. Open pull requests in the
  specification repository are not treated as authoritative.
- Error budgets exposed in YAML, reported externally in #475.
- Upgrade-path criteria from #475 are being counted in the release ledger rather than tracked
  separately.

### Waiting on a maintainer decision

CodeQL alerts #90/#91 (plain-http backend URLs carrying credentials) need a ruling between
refusing them outright and carving out loopback. Both are defensible and the choice changes
what users can configure, so it is not one to make silently.

MikkoParkkola:
### Branch does not compile: a field landed on one side only

`b7a4a768` (input bridge into the invoke path) is on this branch and reads
`self.policy_epoch` at `src/gateway/meta_mcp/invoke.rs:1080`. The committed
`src/gateway/meta_mcp/mod.rs` has zero occurrences of that name — the field,
its initialiser and its accessor exist only in the working tree, uncommitted.

Result at head `8b036191`: `error[E0609]: no field policy_epoch on type &MetaMcp`,
which fails Clippy, Tests, Windows check, Kani, Public claims and both smoke jobs.
Nothing downstream of a compile can go green until the declaration lands.

Three things the owning lanes need to do, in this order:

1. `src/gateway/meta_mcp/invoke.rs` — widen `OutboundRetry` to `pub(super)`.
   The working tree currently fails with `error: type OutboundRetry is private`
   at `src/gateway/meta_mcp/mod.rs:2088` (twice), so committing `mod.rs` alone
   trades one compile error for another.
2. Commit `mod.rs` and `invoke.rs` together — the declaration and its use are
   one contract and neither half compiles without the other.
3. Stamp `src/gateway/meta_mcp/policy_epoch_tests.rs` before committing it.
   It is untracked and has no SPDX header, so it will fail the per-file license
   boundary check the moment it appears. `bash scripts/ci/apply-license-headers.sh --apply`.

Fixed on the branch so far: the unclosed delimiter in `src/transport/stdio.rs`
(`8b036191`) and the two test files missing SPDX headers (`0ed90b05`).

PR #160, opened 2026-04-27
fix: resolve CI failures from Rust 1.95 clippy/rustfmt lints

v4de:
Fixes all CI failures (clippy, rustfmt, compilation) introduced after PR 159 merged and the CI runners upgraded to Rust 1.95.

Compilation fixes:
- Add missing structured_content: None to ToolsCallResult initializers in backend.rs and provider/mod.rs

Rustfmt fixes (15+ sites):
- Wrap long function signatures, format! macros, obj.insert calls
- Fix collapsed if-let brace placement
- Reorder use statements and module declarations alphabetically
- Fix line break placement on method chains and closures
- Wrap long .expect() chains in tests

Clippy fixes (Rust 1.95 new lints):
- derivable_impls: derive Default for ResponseContractConfig
- map_unwrap_or / map_or_else: use idiomatic Option combinators
- too_many_lines: allow on build_meta_mcp
- collapsible_if: collapse 3 nested if blocks
- io_other_error: use io::Error::other() (5 sites)
- if_not_else: flip shared_secret.is_empty() condition
- redundant_closure / redundant_closure_for_method_calls: simplify closures
- uninlined_format_args: inline variables in format strings
- double_ended_iterator_last / filter_next: use .rfind()
- doc_markdown: backtick technical terms in doc comments

All changes are mechanical with no logic changes.

PR #474, opened 2026-09-04
test(meta_mcp): bind #465 timing assertions to the timeouts they test; credit @terafin

MikkoParkkola:
Follow-up to #465, which merged before these two items landed.

## Test bounds

An independent review of the #465 diff found that the three timing assertions bound at 10 seconds while the aggregation timeouts they exercise are 100ms and 1s. The backends in those tests carry their own 5-second timeout, so the pre-parallel serial path would also have finished inside 10 seconds — the assertions could not fail for the reason they exist.

They are now 2s and 3s, below that 5-second serial bound and with room for slow CI, and each failure message names the bound it discriminates. No production code changed.

Verified: `cargo test --lib meta_mcp` → 87 passed, 0 failed, in 2.38s.

## Contributor credit

`AUTHORSHIP.md` recorded that the project had no contributors besides the maintainer. That was untrue, so the declaration now points at a `Contributors` section and the three changelog entries name @terafin.

**It had been untrue since April, not since #465.** A review of the first draft of this PR caught the table asserting #465 was the first outside contribution. GitHub's contributor list has five entries: #159 (Bryan Zick, submitted by @v4de, `structuredContent` for tools declaring an `outputSchema`) merged 2026-04-27 and #160 (@v4de, Rust 1.95 lint and compilation fixes) merged 2026-04-28. Both are now in the table, in merge order. Both predate `CLA.md`, which was added 2026-07-11, so their CLA column says `not applicable` rather than `outstanding` — a contribution accepted before the agreement existed was not withheld from it.

**One thing the table records rather than papers over.** This repository is dual-licensed and `CONTRIBUTING.md` asks first-time contributors for a one-line CLA statement in the PR description. #465 has no such line and no sign-off trailer, and it merged without one. The `CLA on record` column says `outstanding` instead of implying an acceptance that did not happen. Asked for on the PR thread; the column flips when it arrives.

## Acceptance criteria

- [x] Each timing assertion fails on the serial path it exists to rule out
- [x] Each failure message names the timeout it discriminates
- [x] `AUTHORSHIP.md` no longer asserts sole authorship
- [x] CLA status stated accurately rather than assumed
- [x] Every outside contribution on `main` appears in the table, not only the most recent
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib meta_mcp` clean

### DIFF
--- a/AUTHORSHIP.md
+++ b/AUTHORSHIP.md
@@ -3,12 +3,16 @@
 **Work:** mcp-gateway (this repository)
 **Author:** Mikko Parkkola, Finland
 **Date of declaration:** 2026-07-12
+**Last amended:** 2026-09-04 (first outside contribution accepted)
 
 ## Declaration
 
-Mikko Parkkola is the sole human author of mcp-gateway. As of this date there
-are no other human contributors. Where future contributions are accepted, they
-are governed by [`CLA.md`](CLA.md).
+Mikko Parkkola is the principal human author of mcp-gateway and the author of
+its protectable expression as described below. Outside contributions accepted
+into the work are listed under [Contributors](#contributors), together with
+whether that contributor's [`CLA.md`](CLA.md) acceptance is on record. Under
+that agreement a contributor grants the licences this project's licensing
+scheme requires while retaining copyright in their own contribution.
 
 The work is an original work of authorship. Its protectable expression — the
 architecture, module structure, control and data flow, interface design, naming,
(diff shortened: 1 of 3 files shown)


MikkoParkkola:
## Review verdict

Reviewed by GPT-5.x (codex): **SHIP-WITH-FIXES** — "the chain-of-title update omits an earlier outside contributor".

That finding was correct and is now fixed in d597c0b. Disposition of all four items:

| Finding | Disposition |
|---|---|
| `AUTHORSHIP.md:6` claims #465 was the first outside contribution; earlier commits exist | **Fixed.** Verified against the repository's own contributor list, which has five entries. #159 and #160 merged in April and are now in the table in merge order. |
| #465's CLA is still not on record | **Recorded, not fixable here.** The table says `outstanding` precisely because it is; the request is on the #465 thread and the column flips when it arrives. Gated before release, not before merge. |
| Use the 100ms aggregation timeout in the fast-plus-hung test instead of 1s | **Declined, with reason.** It saves ~0.9s of a 2.38s suite and buys a 100ms wall-clock budget for a real HTTP round-trip on contended CI. The assertion's discriminating power against the 5s serial path is identical either way, so the trade is flake risk for nothing that changes what the test proves. |
| State the release and deployment stage | **Noted.** This is a follow-up to a merged PR on `main`, ahead of the 4.0.0 release; no separate deployment stage applies. |

The two pre-#465 contributions turned out to predate `CLA.md` (added 2026-07-11) by about ten weeks, so their CLA column reads `not applicable` rather than `outstanding` — a contribution accepted before the agreement existed was not withheld from it.

Issue #462, opened 2026-08-29
An unparseable config is replaced with defaults, then written back to disk


MikkoParkkola:
## Problem

`load_config_or_default` turns any `Config::load` failure into `Config::default()` after a
warn-level log (`src/config_persistence.rs:13-22`). One of its callers is the admin-UI
read-modify-write path, which then writes the result straight back to disk:

```rust
let mut config = crate::config_persistence::load_config_or_default(path);   // :1405
let value = match mutate(&mut config) { ... };
crate::config_persistence::write_config(path, &config)?;                    // :1410
```

(`src/config_reload/mod.rs:1405-1410`)

So a YAML syntax error in a config an operator is editing by hand does not surface as an
error at the admin UI. The gateway loads defaults, applies the requested mutation to those
defaults, and overwrites the file — every backend, credential reference and policy in it
replaced by a default config carrying one edit. The only trace is a `Could not load config,
using defaults` line in the log.

The CLI paths do the same on `add`/`remove` (`src/commands/add_remove.rs:71,136,241`,
`src/commands/setup.rs:77,418`), where defaulting on a MISSING file is the intended
behaviour — the two cases are conflated in one function.

`load_existing_or_default` (`src/config_persistence.rs:29-35`) already has the shape a fix
would take: default when the file is absent, propagate the error when it exists and cannot
be parsed. Adjacent to #437, which reports data loss from a different config-read failure.

## Acceptance criteria

- [ ] `MCPGW.CFGDEF.1` Given a config file that exists and fails to parse, When an admin-UI
      mutation runs against it, Then the mutation returns an error and the file on disk is
      byte-identical afterwards.
- [ ] `MCPGW.CFGDEF.2` Given a config file that does not exist, When a CLI `add`/`setup`
      runs, Then a default config is still created — the current behaviour, unchanged.
- [ ] `MCPGW.CFGDEF.3` A test asserts the distinction directly: absent file defaults,
      present-but-unparseable file errors, on the same call path.

## Fail-fast

Cheapest check first, before any refactor: write a test that puts a single unbalanced quote
in a config with one backend, calls the admin-UI mutation path, and asserts the file still
contains the backend afterwards. If it already passes, the read-modify-write is protected
somewhere this issue has not found and the scope collapses to the CLI paths.

## ROI

Cost: one fallible sibling plus call-site updates at the two gateway readers; the five CLI
callers keep the current function. Value: removes a path where an editing mistake in a
[...]

PR #483, opened 2026-09-06
fix(config): preserve existing configuration on load errors for 4.0

MikkoParkkola:
Existing malformed, unreadable or dangling-symlink configs can currently fall back to defaults before a mutation writes them out. This increment refuses those edits before changing the file or live state. Missing configs still bootstrap, and valid edits preserve literal secret references.

Targets the 4.0 integration branch because it depends on its literal-config loader. Addresses #462; issue closure remains pending final integration.

Validation:
- Isolated acceptance suite: 44/44 passed, including actual permission denial, setup/client-file preservation and literal references.
- All-feature/all-target Clippy with warnings denied, formatting and diff checks passed.
- Retained evidence: both code reviewers approved the behavioral implementation; independent public drive passed 92/92 assertions; critical guard coverage 29/29 lines and mutation score 8/9 viable caught.

Both reviewers have now approved the subsequent octal-permission spelling and base-helper formatting changes at commit `35be02836e1b054831b7d1db1063cd7794b42b2d` (actual exit 0; material-bound receipts verified). No runtime behavior changed in those deltas. Draft pending exact-commit CI: the current CI workflow triggers only for main pushes/PR targets, so no quality workflow ran for this stacked integration PR. The skipped auto-merge check is not validation evidence. Full evidence and limitations are in `docs/design/issue-462-config-preservation.md`.

This is one delivery increment. It does not establish full 4.0 release readiness, merge or deployment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [main]
     tags: ["v*"]
   pull_request:
-    branches: [main]
+    branches: [main, codex/v4-release-delivery]
 
 permissions:
   contents: read
--- a/docs/design/issue-462-config-preservation.md
+++ b/docs/design/issue-462-config-preservation.md
@@ -1,0 +1,692 @@
+# Issue 462: preserve an existing config when loading fails
+
+## Increment delivery checkpoint — 2026-09-07
+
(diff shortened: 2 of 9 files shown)


MikkoParkkola:

branch holds no unique work.

staleness.

Issue #452, opened 2026-08-28
DELETE /mcp terminates a session without checking the caller owns it


MikkoParkkola:
Reported by an independent review of PR #450 and confirmed at source. Out of scope for that PR: the handler is untouched by it.

**Problem.** `DELETE /mcp` at `src/gateway/router/handlers.rs:249` removes the session named by the `mcp-session-id` header without resolving the caller principal, so it never applies the owner binding that `GET` and `POST` enforce via `session_owner` (`src/gateway/router/handlers.rs:135`). A caller who knows another session's id can terminate that session's stream.

Session ids are UUIDv4 and are not published, which is why this is filed rather than blocking. It is nevertheless the same class of defect PR #450 closes on the dispatch path, left open on the teardown path.

**Acceptance criteria**
- [ ] MIK.SESSION.1 — Given a session created by principal A, When principal B issues `DELETE /mcp` with that session id, Then the session is not removed and the response does not confirm deletion.
- [ ] MIK.SESSION.2 — Given a session created by principal A, When A issues the same request, Then the session is removed as today.
- [ ] MIK.SESSION.3 — Given an unknown session id, When any principal deletes it, Then the response is identical to MIK.SESSION.1 so the endpoint does not confirm which ids exist.
- [ ] MIK.SESSION.4 — A router-level test covers all three branches.

**Fail-fast.** Call the same `session_owner` resolution the GET/POST paths use and compare it with the session's recorded owner before removal. If the multiplexer does not already record an owner per session, that is the real finding and the ticket stops to say so.

**ROI.** Removes an unauthenticated denial-of-service against another tenant's stream and makes session teardown consistent with session use. Cost: small — one lookup and one comparison, plus tests.

**Source.** `src/gateway/router/handlers.rs:249`, `src/gateway/router/handlers.rs:135`.

MikkoParkkola:
Scheduled for 4.0.0, batch 3 (caller identity). Grouped with MIK-7252, MIK-7251, MIK-7257 and MIK-7250, which all turn on proving who a request is for.

MikkoParkkola:
This branch (`feat/mcp-2026-protocol`, the 4.0.0 release train) does not touch this issue. It was reviewed during the 4.0.0 backlog triage and left for a later release; nothing here is fixed, and no code on the branch changes the behaviour described above.

Recorded so the issue is not read as covered by the 4.0.0 work.

MikkoParkkola:
[...]

PR #484, opened 2026-09-06
fix(session): enforce ownership for legacy MCP termination

MikkoParkkola:
Legacy `DELETE /mcp` currently lets a caller terminate another caller's session if they know its ID. This increment checks the validated owner and removes the session atomically. Foreign and unknown IDs return the same 404; owner deletion and auth-disabled anonymous sessions remain supported. Auth-enabled public paths require a valid credential before session lookup.

Refs #452. This is a separate 4.0 stability increment targeting `codex/v4-release-delivery`.

Validation: fresh isolated build passes 11 acceptance tests and 14 existing streaming/ownership controls. All-feature/all-target Clippy with warnings denied, formatting, actionlint and diff checks pass. The unchanged runtime patch retains both SHIP code reviews, 56/56 covered source regions, 10/11 caught viable mutations and seven passing independent endpoint criteria. The test fixture initialization was adjusted for Clippy; the integration-branch CI selector and formatter-only discovery helper change reuse the already reviewed configuration increment.

Source bindings, commands and remaining gates are committed in `docs/release/gh452-session-ownership-evidence.json`; the design documents the acceptance cases and delivery checkpoint.

Draft: mechanical delivery-delta confirmation and CI/integration are still pending. The integration base has two independently reproduced ORDER.2 failures being repaired in a separate prerequisite. No release or issue-closure claim.

## KPI Impact

Cross-caller session termination changes from successful deletion to a non-disclosing refusal, with the original session and stream remaining usable. The independent drive verified all seven acceptance criteria; no numerical performance claim is made.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [main]
     tags: ["v*"]
   pull_request:
-    branches: [main]
+    branches: [main, codex/v4-release-delivery]
 
 permissions:
   contents: read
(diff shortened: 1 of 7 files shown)


MikkoParkkola:

branch holds no unique work.

staleness.

PR #485, opened 2026-09-06
fix(mtls): generate certificates accepted by strict clients

MikkoParkkola:
Gateway-generated certificates are rejected by strict X.509 clients because the CA lacks signing key usages and issued leaves lack an issuer identifier. Set those existing rcgen options for new certificates and document a rollout that retains old trust and supports rollback.

Validation: 5 DER regression tests plus 67 existing mTLS tests pass in a fresh isolated build; all-feature/all-target Clippy with warnings denied, formatting and actionlint pass. Both final code reviewers approved the byte-identical runtime and tests. Both changed executable lines are covered; both viable generated mutants and four targeted faults are caught (two generated mutants were unviable). Independent strict mTLS and isolated rollout/rollback checks pass. Existing keys and trust are unchanged.

The CI branch selector and formatter-only discovery helper match the already reviewed configuration increment. Full CI and integration are pending; the base has two confirmed ORDER.2 failures being repaired in a prerequisite. This PR does not claim the separate HTTP shutdown repair or complete 4.0 readiness.

Refs MIK-7212.TLSCERT.1–4. Evidence and operator instructions are included in the design document and deployment guide.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [main]
     tags: ["v*"]
   pull_request:
-    branches: [main]
+    branches: [main, codex/v4-release-delivery]
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


MikkoParkkola:

branch holds no unique work.

staleness.

PR #486, opened 2026-09-06
fix(protocol): isolate sessionless workflow state

MikkoParkkola:
Sessionless `gateway_set_state` calls previously changed the shared empty-key state, altering discovery for modern HTTP callers. This change returns the ratified protocol refusal and makes all four discovery readers ignore existing empty-key state. Nonempty legacy and stdio owners retain their state transitions.

This is the bounded ORDER.2 prerequisite for the 4.0 integration branch. It also enables the existing CI jobs for that exact integration target. The design and test plan preserve the original findings and approved acceptance criteria.

Validation at `d95d0a9640772207d40f8d5fce4d59243a2e943c`:

- Eight acceptance/caller tests pass after five compiled assertion failures and three controls on the pre-guard source. Modern HTTP is exercised with and without an offered session header; the stdio dispatcher test shares the production owner constant and retains an independent literal assertion.
- Default-feature library: 4,021 pass, zero fail, four ignored. All-feature CI library: 4,059 pass, zero fail, four ignored; binary tests: 222 pass.
- All-feature CLI build, all-feature/all-target Clippy with `-D warnings`, and formatting pass.
- Critical guard coverage: 36/37 regions (97.30%); every changed executable guard region is covered. The uncovered region is the existing missing-argument error branch. No OS-pipe coverage is claimed for the mechanical stdio constant extraction.
- `cargo-mutants`: three generated mutations compiled and were caught, zero missed/unviable/timeouts. Separate write-only and read-only guard omissions caused the intended assertion failures; restoring the original bytes restored all eight passing tests.
- Design/test-plan, tests-as-tests, and final GPT/Grok code gates are closed. Final reviews are both SHIP with actual exits 0, identical material SHA `bb57b4c36f0bb42870789f88bba3d1ae49de59168a6d9c397fa77b0e649ebcdd` (28,847 bytes), bound to this commit.

**Merge remains blocked.** [CI run 34066309790](https://github.com/MikkoParkkola/mcp-gateway/actions/runs/34066309790) passes the repaired B08/B09 library cases, then fails the unchanged TASK.1 integration case `tasks::ac_task_1_a_failed_task_reports_its_failure_rather_than_an_absence`: the base returns a string instead of a structured JSON-RPC task error. TASK.1 is a separate release work package. Independent public acceptance now passes on this same pinned executable: one isolated Spark launch, 40 recorded requests, exact fixture readiness, all four discovery surfaces, both modern refusal variants, and two distinct legacy sessions with state-dependent isolation. Internal None/contaminated-state/stdio clauses are mapped to the separately passing component tests; they are not presented as black-box observations. The original failed readiness run is retained. No full Task or release acceptance is inferred. This PR remains draft; no release or ticket closure is claimed.

## KPI Impact

Restores a stable default discovery set for sessionless callers. The eight-case suite covers `MIK-7272.ORDER2.FSM.1` through `.5`, with internal-only observations explicitly separated from public functional acceptance.

Tracks [MIK-7272](https://linear.app/parm/issue/MIK-7272).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [main]
     tags: ["v*"]
   pull_request:
-    branches: [main]
+    branches: [main, codex/v4-release-delivery]
 
 permissions:
   contents: read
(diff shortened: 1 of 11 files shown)


MikkoParkkola:

branch holds no unique work.

staleness.

PR #515, opened 2026-09-08
fix(lifecycle): report reclaimed session count

MikkoParkkola:
The CONTROL.4 count acceptance tests cannot compile because SessionLifecycle::reap returns (). Return the number of expired keys removed while preserving existing callback and lock behavior.

Validation: both count acceptance tests and all six lifecycle tests pass on a forced-rebuild Spark snapshot; formatting and diff checks pass. Independent review pending. This addresses CONTROL.4 T3 and the type mismatch only; cleanup wiring and the full release criterion remain open.


MikkoParkkola:
Superseded by #516, which carries this change plus the idle-deadline sweep that gives the count something to report. The branch is kept; nothing here is lost.

PR #516, opened 2026-09-08
feat(session-lifecycle): reclaim per-identity state on an idle deadline

MikkoParkkola:
## What

MCP 2026-07-28 removed protocol sessions. Nothing fires a disconnect any more, so every
per-session cleanup handler would simply never run and everything it reclaimed would leak
in silence — nothing errors when a callback is not called.

This replaces the disconnect trigger with an idle deadline.

- Track each identity with an expiry; one deadline per key, replaced on refresh rather than accumulated
- Sweep expired deadlines on the host reaper tick, firing each key's handlers exactly once
- Return the reclaimed count from `reap` so a sweep can log what it took
- Mark every sweep, so an empty sweep is observable rather than indistinguishable from no sweep

Closes MIK-7215 CONTROL.4.

## Evidence

7 tests pass across 4 acceptance binaries (`mik_7215_control4_{track,wiring,reap_count}_acs`,
`mik_7215_control4_sweep_log`); suite exit 0.

## Known gaps, stated rather than implied

- A mutation probe broke `session_lifecycle.rs` at two points and the suite stayed green.
  The acceptance tests bind the wiring and the sweep log, not every branch of the reaper.
- `fire_cleanup` carries a stated residual: a key re-tracked between the sweep's removal and
  the handler call still has its handlers fired. Closing it needs an ownership model this
  module does not have, and the path is not reached from production (MIK-7291).
- Second-opinion reviews were recorded one commit before this tip.

## Supersedes

Replaces #515, which covered only the reclaimed-count change.


MikkoParkkola:
Closing as superseded — the CONTROL.4 reaper is already on the release line.

The commit counts make this hard to see. This branch forks from `main` at `c3626cf8`, the same point the `codex/v4-*` lineage forks at, so it reports 1,699 commits ahead of `main`. Only **6** of those are its own work: `origin/codex/v4-next-integration` is 270 commits ahead of it, and it is 6 ahead of codex. Those 6 are the MIK-7215 CONTROL.4 idle-session reaper.

That feature is already landed. All four acceptance suites exist on `main` and on `feat/sub2b-outbound-mint`, and each is byte-identical to the copy on this branch (compared by blob SHA, 2026-09-11):

- `tests/mik_7215_control4_reap_count_acs.rs`
- `tests/mik_7215_control4_sweep_log.rs`
- `tests/mik_7215_control4_track_acs.rs`
- `tests/mik_7215_control4_wiring_acs.rs`

`MIK-7215.CONTROL.4` reads MET in the release criteria ledger accordingly. Merging this branch would carry the entire codex lineage through a PR whose title names a session-lifecycle reaper, for a feature that is already present.

Re-derive with `git rev-parse <branch>:tests/mik_7215_control4_reap_count_acs.rs` on both branches. Evidence recorded in `docs/requirements/RELEASE-4.0.0-gap-assessment-2026-09-11.md`.

Issue #517, opened 2026-09-10
streamable_http backend: initialize lacks protocol version negotiation fallback (400 vs servers supporting <= 2025-06-18)


luochen1990:
## Environment

- mcp-gateway 3.5.0 (nixpkgs build)
- Backend: `https://mcp.docs.astro.build/mcp` (official Astro docs server, `streamable_http: true`)

## Symptom

The backend never becomes healthy. Health probe fails in a tight loop (every ~10s, each cycle logs `Health probe failed; rebuilding transport` → `Starting backend transport` → fail again):

```
WARN mcp_gateway::gateway::server: Health check failed backend=astro-docs error=Transport error: HTTP 400 Bad Request: {"jsonrpc":"2.0","error":{"code":-32000,"message":"Bad Request: Unsupported protocol version (supported versions: 2025-06-18, 2025-03-26, 2024-11-05, 2024-10-07)"},"id":null}
```

## Analysis

The streamable HTTP client sends the latest spec version (2025-11-25, per #73) in `initialize`. The Astro server only supports up to **2025-06-18** and rejects with HTTP 400, listing its supported versions in the error body.

Per the MCP spec's version negotiation, when a server responds with 400 + supported versions, the client **should retry with a mutually supported version** instead of failing permanently. The gateway appears to have no such fallback, so any backend server that is a few spec revisions behind can never connect.

Notably `https://mcp.docs.astro.build/mcp` works fine when driven directly with `initialize (2025-06-18) → notifications/initialized → tools/list` (it's stateless — no session header returned), so this is purely the negotiation gap on the gateway side.

## Suggested fix

On `initialize` failure with 400 + `supported versions` in the body, parse the list, pick the newest mutually supported version (fall back through 2025-06-18 / 2025-03-26 / 2024-11-05), and retry. Persist the negotiated version per backend for the transport's lifetime.

A related observation (happy to file separately if preferred): `/health` reports `backends.all_healthy: true, count: 8` while this backend's probe was failing continuously — the aggregate seems not to account for the failing backend.

Thanks for the great gateway — the multi-backend + auth-key model has been working very well for us.

luochen1990:
Update after testing: **v3.5.1 does not fix this** (ran the v3.5.1 linux binary against `mcp.docs.astro.build` with our config — same tight failure loop, no "retrying with negotiated version" log line).

## Refined root cause

The negotiation fallback added in v3.5.1 only triggers on a **JSON-RPC error response**:

[...]

PR #521, opened 2026-09-10
fix(idempotency): release a provably pre-dispatch failure and bind the key to its caller

MikkoParkkola:
## Scope

Release-4.0.0 readiness work on the MIK-7272 revision surface: the idempotency
release-versus-settle decision, the identity binding of the idempotency key, and
the criteria-ledger bookkeeping that scores them.

## What lands

- `SUB.4` — a transport failure that provably never left the origin releases the
  idempotency key; every other shape settles terminally. The evidence is a
  redirect counter sampled either side of `send()`, not the error's URL, which
  reqwest back-fills with the original request URL on the error path and so
  cannot distinguish a followed hop. `Error::TransportConnect` via
  `safe_request_error_for` (`src/security/http_diagnostics.rs:88`).
- `SUB.4`'s prerequisite — the idempotency key now binds the calling principal.
  The client-supplied key is length-prefixed so a caller cannot forge another
  caller's key by embedding a delimiter.
- Criteria ledger, blocking rollup and readiness board updated to match source.

## Verification

`cargo test --all-targets`: 4985 passed, 0 failed across 65 binaries.
`cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

## Known red: `tests/mik_7272_sub2b_acs.rs`

Three acceptance rows for `SUB.2b` are committed and failing by design — they
were written first and the implementation is not in this branch. CI will be red
on them. They are the last blocking row in the release ledger
(`146 criteria, 186 rows, 185 met or non-blocking, 1 blocking`).

`SUB.2b` needs three separate pieces, documented in
`docs/requirements/RELEASE-4.0.0-blocking-rollup.md`:

1. The SSE body is read whole by `response.text().await`
   (`src/transport/http/mod.rs:1373`), so a notification cannot be emitted
   before the result it must precede. A streaming read replaces it.
2. The stdio serve loop awaits each dispatch inline
   (`src/gateway/server/mod.rs:1696`, `dispatch_single_with_sink` at `:1737`),
   so a second call cannot be read while the first is in flight.
3. No per-request sink carries a notification to the client ahead of the
   response.

Draft until those land.


MikkoParkkola:
Correction to the "Known red: `tests/mik_7272_sub2b_acs.rs`" section above: `SUB.2b` needs four pieces, not three. In addition to the streaming-read, concurrent-dispatch, and per-request-sink items listed, the sink itself does not exist yet at HEAD (`rg -n 'notification_sink|NotificationSink' src/` returns nothing) and needs to land first — a `tokio::task_local!` mirroring the existing `TRACE_ID` pattern, not just the bounded channel and writer task around it.

A design review of the SSE-read approach (two independent reviewers) also surfaced issues in `establish_sse_connection` (`src/transport/http/mod.rs:1089-1140`) that any implementation should account for: per-chunk `from_utf8_lossy` can corrupt a multi-byte codepoint split across chunk boundaries, the 64 KiB size cap is checked before the complete-line drain and is sized for the handshake case only, and the buffer remainder is reallocated per line (quadratic on line-dense input). Separately, `notifications/message` isolation-by-stream over HTTP has no acceptance-test coverage yet — the existing `SUB.2b` rows are all stdio.

Unrelated to the above: this run's "Public repo hygiene" check failed on `FAIL: cargo package tarball would omit LICENSE-MIT` after a `printf: write error: Broken pipe` in `scripts/ci/verify-artifact-licenses.sh`. `cargo package --list --allow-dirty` run locally on this branch does contain `LICENSE-MIT`, so this looks like a flake in that script's pipeline (not a real packaging omission) rather than something introduced by this branch — flagging in case it recurs on re-run.

MikkoParkkola:
Containment evidence. This branch is fully contained in #528 (`feat/sub2b-outbound-mint`), by
three independent measures rather than ancestry alone:

| Check | Command | Result |
|---|---|---|
| Commits ahead | `git rev-list --count origin/feat/sub2b-outbound-mint..origin/fix/gh517-protocol-negotiation` | **0** |
| Content difference, all paths | `git diff --stat origin/feat/sub2b-outbound-mint...origin/fix/gh517-protocol-negotiation` | **empty** |
| Ancestry | `git merge-base --is-ancestor` | head `ab9d1aaa` is an ancestor of #528 |

The content check is the one that matters: a zero commit count can coexist with a divergent
tree, and it does not here. There is no file, in `src/`, `tests/` or anywhere else, that this
branch carries and #528 does not.

Nothing is lost by closing this PR — the work ships through #528. Leaving the disposition to the
repository owner rather than closing it here.

MikkoParkkola:
Closing as fully contained in #528.

`git rev-list --count origin/feat/sub2b-outbound-mint..origin/fix/gh517-protocol-negotiation` returns **0** — this branch has no commit that the release line does not already carry. The diff between them runs the other way: #528 is ahead by 307 insertions and 10,257 deletions across 55 files.

GH #517 itself landed separately as #520. Nothing here is lost by closing; if a specific hunk is wanted it is already on `feat/sub2b-outbound-mint`.

Evidence recorded in `docs/requirements/RELEASE-4.0.0-gap-assessment-2026-09-11.md`.

PR #522, opened 2026-09-11
fix(windows): pass APPDATA and LOCALAPPDATA to stdio children

yfcyfc123234:
## Summary
- On Windows, `configure_child_environment` clears the child env and only re-injects a small allowlist (`USERPROFILE`, `TEMP`, `TMP`, `SYSTEMROOT`, …).
- Python’s **user** site-packages live under `%APPDATA%\Python\…`. Without `APPDATA` / `LOCALAPPDATA`, stdio backends that rely on user-site installs (e.g. `fastmcp` via `pip install --user`) fail at import with `ModuleNotFoundError`, and the stdio transport then surfaces a **request timeout** instead of a clear spawn/import error.
- This PR adds `APPDATA` and `LOCALAPPDATA` to the Windows allowlist next to `USERPROFILE`.

## Repro (before)
1. Install an MCP server dependency into the user site (`pip install --user fastmcp` / typical Windows Python layout).
2. Configure it as a stdio backend and start the gateway.
3. First `tools/list` / initialize against that backend times out; child stderr shows missing modules from user site.

## Test plan
- [ ] `cargo test -p mcp-gateway transport::stdio` (or full suite if preferred)
- [ ] On Windows: stdio backend whose deps live only in `%APPDATA%\Python\…\site-packages` initializes and returns `tools/list`
- [ ] Confirm secrets still do not leak: child env remains cleared aside from the allowlist + per-backend `env:`

## Notes
- Minimal two-line change; no behavior change on non-Windows.
- Separately we hit Windows `shlex` backslash path issues and FastMCP `Errno 22` on Tokio pipes — happy to file those as follow-up issues if useful.

### DIFF
--- a/src/transport/stdio.rs
+++ b/src/transport/stdio.rs
@@ -56,6 +56,8 @@ fn configure_child_environment(cmd: &mut Command, backend_env: &HashMap<String, 
     #[cfg(windows)]
     for key in [
         "USERPROFILE",
+        "APPDATA",
+        "LOCALAPPDATA",
         "TEMP",
         "TMP",
         "SYSTEMROOT",


MikkoParkkola (approved):
Reviewed against the code at `main`. Both the defect and the fix hold up, and the fix is the right size. Merging this.

**The defect is real.** `configure_child_environment` (`src/transport/stdio.rs:41`) calls `env_clear()` and re-injects an allowlist that has `USERPROFILE` but not `APPDATA` / `LOCALAPPDATA`. Python resolves its user site-packages from `%APPDATA%`, so a backend installed with `pip install --user` cannot import its own dependencies.

**It is wider than Python.** npm's global prefix on Windows is `%APPDATA%\npm`. Any `npx`-launched stdio backend relying on a global install hits the same wall, which covers a large share of published MCP servers.

**The security posture is unchanged.** `USERPROFILE` is already forwarded, and `%APPDATA%` is `%USERPROFILE%\AppData\Roaming` by definition — the child could already derive the path. An environment variable conveys a path, not access; the file system decides what the child may read. `env_clear()` plus an allowlist stays intact, and the per-backend `env:` override still wins.

**On the missing test.** The one test covering this code, `backend_subprocess_receives_only_safe_and_explicit_environment` (`src/transport/stdio.rs:987`), is `#[cfg(unix)]`, and the Windows CI job runs `cargo check --all-features` — it compiles the crate and never executes a test. Asking for a test here would mean asking for a test that nothing runs. That gap is the maintainer's to close, not a condition on this PR. Follow-ups filed on our side:

- Run the test suite on the Windows runner, not just `cargo check`.
- Cover the Windows branch of the allowlist with a test that asserts both what is forwarded and what is not.
- Review the allowlist as a whole. It is a denylist by omission, so the next absent variable is the next timeout report; `PROGRAMDATA`, `PROGRAMFILES`, `SystemDrive`, `windir`, `HOMEDRIVE`/`HOMEPATH` and `USERNAME` are the candidates.

**The most valuable thing in this report is not the two lines.** A child that dies on `ModuleNotFoundError` surfaces to the operator as a request timeout, with the actual cause sitting unread in the child's stderr. That is why this took a user-side repro to find at all, and it is getting its own issue.

**Next step here:** the required workflows were sitting at `action_required`, as fork PRs do until a maintainer releases them. Released now; the merge follows once CI is green.

Thank you for the precise report, and yes please on the follow-ups you offered — #523 is being answered separately.

MikkoParkkola:
Follow-ups now filed and linked: #524 (Windows CI runs a test), #525 (allowlist test, both halves), #526 (a child that dies before initialize reports its stderr rather than a timeout). The parser work from your second report is #527.

PR #528, opened 2026-09-11
feat(release): close the v4.0.0 request-scoped notification and outbound refusal gaps

MikkoParkkola:
## Scope

Closes the remaining v4.0.0 release-criteria gaps on the request-scoped
notification path (MIK-7272.SUB.2b) and the outbound direct-route refusal
path, plus the supporting transport and firewall corrections found while
auditing the criteria ledger.

Tracked against `docs/requirements/RELEASE-4.0.0-criteria-status.md`.

## What changed

- **Outbound direct route** — an era-removed method is refused with the
  caller's request id instead of a synthesised one; outbound health probes
  are gated on the peer's protocol era.
- **Transport** — a JSON-RPC error carried on a retryable HTTP status keeps
  the peer's code for every reader (`to_rpc_code`, session-expiry detection,
  era classification); session recovery resends the modern liveness probe.
- **Inbound notification capture** — `parse_sse_response` returns the
  notifications observed alongside the response rather than dropping them;
  stdio captures by progress-token match.
- **Security** — drift detection between merged and listening controls;
  the superseded blocked-response payload builder is removed in favour of
  the `-32600` refusal the router already serves.

## Verification

- `cargo test --lib` — 4221 passed, 0 failed, 3 ignored
- `cargo clippy --all-targets -- -D warnings` — clean
- Independent review: grok `SHIP-WITH-FIXES` (both findings fixed and
  pinned by test), kimi `SHIP`

## Out of scope

- Extracting a shared retryable-status classifier (duplicated in
  `src/chains/retry.rs` and `src/failsafe/retry.rs`) — flagged by both
  reviewers, filed rather than folded in.
- `notifications/roots/list_changed` on the direct backend route — a
  different mechanism (no id, no response channel); filed as a follow-up.

## Status

Draft until the MIK-7272.SUB.2b outbound leg lands. Per the standing merge
constraint, the inbound capture scaffold ships with the outbound emitter or
not at all.


MikkoParkkola:
## v4.0.0 DoD evidence — measured at `00017f89`

CI is green on the pushed tip. This records the measurement, and which rows of
`docs/requirements/RELEASE-4.0.0-dod-check.md` it supersedes. That document is a point-in-time
record taken at `c3083368` and is not edited retroactively.

### Suite

| Metric | Value |
|---|---|
| Test binaries | 96 |
| Passed | 5511 |
| Failed | **0** |
| Ignored | 30 |

Source: the `Tests` job of the CI run on `00017f89`, `cargo test --all-features --no-fail-fast`.

### Gate jobs

All 18 jobs of the CI workflow report success, including every job that was red on the previous
pushed tip `e0f9396b`:

| Job | `e0f9396b` | `00017f89` |
|---|---|---|
| Tests | failure — `--test mik_7272_sub2b_acs` | **success** |
| Clippy (pedantic) | failure — 3 errors in `mik_7272_sub2b_acs` | **success** |
| Secret-leak lint (CWE-532) | failure — 2 findings interpolating a token-shaped value | **success** |
| Release criteria ledger | failure — header/row disagreement | **success** |

Also green: Control-drift probes (NFR.SEC.7), Kani, Format, Windows check, Dependency audit,
Secrets scan, Public claims, Public repo hygiene, First-use usability smoke, Service template
smoke, Helm chart lint + render, Helm OCI package/push/pull, Helm air-gap export/import, Helm
supply chain (cosign + SBOM), K8s kind apply+upgrade+rollback.

### Criteria ledger

`python3 scripts/release/count-release-criteria.py` reports
**149 criteria, 189 rows, 188 met or non-blocking, 1 blocking**, and `--check` exits 0.

### The one remaining gate

`NFR.SEC.7` is the sole blocking criterion, and its open half is a deployment rather than a code
change. The control is unconditional in code — no toggle exists — and the control-drift probe
job passes in CI against this tip. What is unmet is the running process: it is
`3.4.0-f30539af`, built before `5d25f104` added the control, so the guard cannot be observed in
force on the live port until that process is replaced.

### Independent review

The last implementation commit on this branch (`e8c941ca`, clearing the declared log level when a
request declares none) was reviewed by two independent non-Claude reviewers. Both returned SHIP.
Three improvements were raised across them; all three are landed — `b727a7c9` and `00017f89`.

### What this supersedes in the `c3083368` dod-check

- **§4 / D1** — the suite was recorded FAIL with 14 failures at that head. It is green here, and
  the `every_cited_test_exists` failure specifically is repaired: the citation at
  `tests/mik_7272_conformance.rs:150` resolves to a test defined at `tests/mik_7212_acs.rs:449`.
- **§12** — the reviewer-availability note is inverted relative to the current state.
- **The `NFR.SEC.7` row** — superseded by the code/deploy split, and by the correction that the
  origin guard is unconditional rather than enabled by default.

The H1–H11, D1–D30 and §1–13 tables otherwise stand.

MikkoParkkola:
## Retro — v4.0.0 release-readiness pass (MIK-7272.SUB.2b)

Recorded against the DoD §13 requirement: what worked, what did not, what is next.

### What worked

- **Measuring instead of asserting.** Coverage was compared as a matched pair — the same command at the merge-base `738c7cee` and at `f903b854`, both excluding the same instrumentation-hostile rows — which turned a leg that had twice been written off as unobtainable into a result: 85.12% to 85.40%, no drop.
- **Deriving counts from the tables rather than from the previous revision's prose.** Every "N rows outstanding" claim in the record is now re-counted off the matrix each time the matrix changes; two earlier counts were wrong precisely because they were carried forward.
- **Reading the gate definition before scoring the gate.** Three rows filed behind a deployment turned out to be readable in the working tree, and scoring them there closed them.

### What did not work

- **Citations in the wrong format broke CI.** `path:line` was written into a JSON field whose contract is a bare repository path, which the acceptance checker resolves with `is_file()`. The fix was the data, not the checker: widening the checker to discard `:190` would also have made it accept `:99999`, and this is the gate whose output says the release is shippable.
- **Verdict prose drifted from the tables it summarised.** Counts, round tallies and change-size figures each disagreed with their source at least once across revisions.
- **"Blocked on deployment" was used as a filing category rather than a finding.** It is accurate for a narrow set — the end-to-end pass, the live half of DAST, rollback execution — and was wrong for everything else it had collected.

### What is next

- A user-acceptance pass. §11 stops the line on a UAT that has never been run, and that condition is independent of the two operator acts.
- A soak configuration in CI. §10 asks for one on a change of this size and asks explicitly that its absence be flagged.
- Latency, token and memory signals for the paths this branch adds. It adds three error-rate counters and no timing.
- Retro and Done Log as a standing step rather than a closing one, so §13 is not assessed for the first time at the end.

MikkoParkkola:
## UAT evidence — v4.0.0 candidate driven as a client would drive it

Independent acceptance driving against the **running binary**, per the FUNCTIONAL PASS gate in
`docs/requirements/RELEASE-4.0.0-dod-check.md` §1. The driver was given the wire surface and the
acceptance behaviours only — no requirements document, no status ledger, and no permission to read
the implementation to decide a verdict.

**Revision under test**: `9c1ab3d7a0cf000c316858068f39b1191db89d02`
**Binary**: `target/debug/mcp-gateway`, `--version` → `4.0.0` (recorded as the report's first evidence line)
**Surface driven**: `POST /mcp` and `GET /mcp` over HTTP on loopback ports 8801–8806, one fresh
gateway + fixture-backend pair per run. Notifications were produced by a stdlib HTTP fixture backend
registered as `fixture` and reached through `gateway_invoke`; the gateway does not manufacture
notifications, so a backend that emits them is the only way to observe where they land.

### Outcomes, against the criteria ledger

| Driven behaviour | Criterion | Outcome |
|---|---|---|
| Request-scoped notifications (`progress` **and** `message`) arrive on the response stream of their own request, carrying the client's own `progressToken` | `MIK-7272.SUB.2b` | **PASS** |
| `notifications/progress` and `notifications/message` are **not** delivered on the subscription stream | `MIK-7272.SUB.2a` | **PASS** |
| Modern path: no standalone SSE stream, no event ids, no `Last-Event-ID` redelivery | `MIK-7272.SUB.3` | **PASS** |
| `-32020` / `-32021` / `-32022` replace `-32001` / `-32003` / `-32004` | `MIK-7272.ERROR.1` | **PASS** |
| Unknown resource URI returns `-32602`, not `-32002` | `MIK-7272.ERROR.2` | **PASS** |
| Foreign `Origin` and foreign `Host` refused; correct ones accepted (positive control included) | *no criterion row* — see note | **PASS** |

Six behaviours driven, six PASS, no unresolved INVESTIGATE.

**On the last row**: the criteria ledger has no DNS-rebinding criterion — `rg -n rebinding
docs/requirements/RELEASE-4.0.0-criteria-status.md` returns nothing. That behaviour was driven
because the DoD's own security rows assert it, and it passed on this revision's binary with both
negative cases and the positive control. It is **not** a statement about any deployed build, which
is what those rows ask for separately.

### The one correction worth recording

The first drive returned **FAIL** on the `notifications/message` half of `MIK-7272.SUB.2b`: the
gateway decoded two notifications from the backend and forwarded one. That conclusion was wrong, and
the defect was in the brief, not the binary — the prescribed request `_meta` omitted
`io.modelcontextprotocol/logLevel`, and a request that declares no level is not owed a
`notifications/message` frame. The driver eliminated two benign explanations before re-driving
(client `logging` capability declared in full: no change; the same drop on the legacy `2025-06-18`
path: not a modern-revision regression), which is why the re-drive was aimed at the request shape.

The re-drive establishes a filter rather than a passthrough, two rows differing in exactly one
`_meta` value:

- `logLevel: "debug"` → the fixture's `{"data":"uat-marker","level":"info","logger":"fixture"}`
  arrives on the POST's own response stream, ahead of the `id` result, alongside the progress frame
  carrying the client's `uat-token-1`. A third frame appears that the first drive never saw: the
  gateway's **own** `notifications/message` at `logger: "gateway.invoke"`.
- `logLevel: "error"` → both `info` messages are withheld; the progress frame, which carries no
  level, is unaffected.

Anyone reproducing this must declare the level or they will reproduce the false FAIL.

### Scope and limits, stated rather than implied

- **Token translation is not distinguished from pass-through.** The observable clause holds — the
  frame carries the client's own token — but the fixture echoes whatever token the gateway sent it,
  so mint-and-translate and pass-through are indistinguishable from the client side. Not resolved.
- **A debug binary on loopback.** This is what *this revision* does when driven. The DoD rows that
  ask for a deployed build (`AC9`, and `NFR.SEC.7`) are unaffected by this evidence and stay open.
- **The report's closing summary is stale.** The AC-1 section was updated to PASS after the re-drive;
  the closing table row and the "one failure" paragraph beneath it still carry the original FAIL and
  were not rewritten. The transcript is authoritative — it contains the raw response bytes for both
  rows — and the inconsistency is recorded here rather than silently resolved.

Full transcript, every command and its verbatim output: kept with the session's evidence, not
committed to the repository.

MikkoParkkola:
## Prerelease channel — v4.0.0 staging path (D21 CANARY)

`v4.0.0-rc.1` now reaches only the channels a user opts into. Seven publish surfaces fire on
a `v*` tag; each was audited and each is now either gated on the channel or explicitly
stable-only.

| Surface | On `v4.0.0-rc.1` |
|---|---|
| GitHub release (`release.yml release`) | Marked prerelease; the "Latest" pointer stays put |
| crates.io (`release.yml publish`) | `4.0.0-rc.1`, unresolvable by `^4` |
| npm (`release.yml npm-publish`) | `@next` dist-tag |
| Homebrew tap (`release.yml homebrew-update`) | Skipped |
| ghcr, signed publisher (`ci.yml docker`) | `:4.0.0-rc.1` only; `:latest` withheld |
| ghcr, metadata publisher (`docker.yml build`) | `:4.0.0-rc.1` only; no `:4.0`, no `:latest` |
| MCP Registry (`docker.yml publish-mcp-registry`) | Skipped |

All three tag-triggered workflows run `scripts/release/check_tag_manifest.py` themselves —
none can read another's job outputs, so a single shared gate job would leave two workflows
ungated. The gate refuses to publish when the tag disagrees with `Cargo.toml`/`Cargo.lock`,
which is what stops `cargo publish` shipping a candidate under the final version number.

Runbook: `docs/release/v4.0.0-prerelease-channel.md`, including the consumer-side verify
snippet — both ghcr publishers push `:VERSION` from the same commit with no ordering between
them, so a consumer verifies by digest against both publisher identities.

### Evidence

- `scripts/release/test_check_tag_manifest.py` — 27 tests: the channel predicate (semver
  build metadata included), the lockfile parse, tag/manifest disagreement in both
  directions, the dispatch-input fallback, and the workflow wiring the gate depends on.
- `scripts/release/test_workflow_wiring_mutations.py` — 14 mutation cases pinning what those
  wiring assertions catch (a deleted `cosign verify`, a step that lost its `DIGEST` binding,
  an identity relaxed to a regexp, a commented-out gate invocation, a `needs: verify` reduced
  to a comment, a prerelease skip deleted, a dispatch tag moved into a `run:` block) and what
  they tolerate (quoted values, folded conditions, trailing comments, block-form `needs`).
  Both suites run in CI.
- Three review rounds on the change; every finding either fixed and mutation-proved or
  recorded with its reason.

### Residual, tracked

#529 — both publishers push first and sign second, so a run that fails between the two steps
leaves `:VERSION` public and unsigned. Detectable (the run is red and the verify snippet
returns non-zero) but not prevented. Closing it means pushing by digest and promoting the tag
only after signing, whose correctness turns on `docker buildx imagetools create` preserving
the signed digest — not the default, and not confirmable without a registry to try it
against.

### Still outstanding

Both remaining OUTSTANDING rows in `docs/requirements/RELEASE-4.0.0-dod-check.md` are owner
acts, not missing work:

- **D21 CANARY** — the mechanism exists and is validated; exposure has not been staged. It
  stays outstanding until an rc is actually published and observed. Promotion criteria and the
  observation window are deliberately left to the owner rather than encoded in the runbook.
- **D18 MERGED** — merge.

PR #531, opened 2026-09-12
chore(personal_accounts): annotate deferred per-user OAuth dead-code cluster

MikkoParkkola:
**Draft. Does NOT make the lint gate green — see "What still fails".**

This PR branch is shared: it carries 13 commits ahead of `codex/v4-next-integration`, of which **only `76659582` is mine**. The other 12 belong to earlier sessions on this branch and I make no claim about them.

## What `76659582` does

`MIK-6744/6745/6746` (per-user OAuth) are deferred to the post-4.0.0 backlog (`docs/requirements/RELEASE-4.0.0-backlog-triage.md:98`) and appear in neither `RELEASE-4.0.0-requirements.md` (73 IDs) nor `RELEASE-4.0.0-criteria-status.md` (80 rows). The consent/grant-commit machinery therefore has no non-test caller *by design*, and the lint gate reads it as dead code.

22 items across 11 warning sites now carry:

```rust
#[cfg_attr(not(test), expect(dead_code, reason = "per-user OAuth scaffolding, deferred to post-4.0.0 backlog MIK-6744/6745/6746"))]
```

- `expect`, not `allow`: the annotation self-deletes the moment production wiring adds a real caller.
- `cfg_attr(not(test), ..)` is load-bearing: lib and lib-test are separate compile units, and the in-tree test callers already satisfy `dead_code` there, so a bare `expect` would itself become an `unfulfilled_lint_expectations` error.

Sites: `mod.rs` (`commit_grant`, `revoke`, `mark_reconnect_required`), `service.rs` (`RuntimeNotImplemented`, `StaleConsentFenced`, `store`, `resolve`, `invalidate`, `commit_grant_if`), `worker.rs` (2x `RuntimeNotImplemented`, `capacity`, `resolve`, `invalidate`, `commit_grant_if`), `commit.rs` (`revoke`, `mark_reconnect_required`), `config.rs` (2 variants), `identity.rs` (2 variants), `consent.rs` (1 variant).

Also corrects `CI-GREEN-PROGRESS.md`, which had inferred the cluster was missing v4 integration wiring rather than deferred scope, and which still described the cluster as an unfixable blocker.

## Verification (real exit codes, on this commit)

| command | before | after |
|---|---|---|
| `clippy --all-features -- -D warnings` (the gate, `ci.yml:180` **and** `release.yml:121`) | **101** — 36 lib + 1 bin errors | **101** — 25 lib errors |
| `test --all-features --no-run` | **0** | **0** |
[...]

PR #512, opened 2026-09-08
Integrate current protocol work with v4 account and security increments

MikkoParkkola:
Combine protocol work through60b138bb with the validated account, task and security increments. Preserve account credential resolution before cache lookup while retaining the updated caller-identity rules. Include offline store initialization, adapter configuration/load-order checks, firewall verdict and multi-target fixes, seven task lint fixes, CI branch coverage and credential Debug evidence. Runtime OpenWebUI middleware remains a separate increment.

Validation: 4,760 library tests pass, zero failures, seven ignored. Formatting, diff and ledger checks pass; ledger remains19blocking.

Blocked: all-target/all-feature check fails on CONTROL.4 reap-count tests because SessionLifecycle::reap still returns(). These intentional red tests remain with the existing Claude CONTROL.4 owner. Full release CI, integration tests, performance/mutation evidence and hosted consent remain incomplete. This is a combined candidate, not release approval.

### Integrated OpenWebUI identity middleware

Commit cc3d2c70 brings the reviewed native signed-header adapter into this integration candidate, including the shared-header multi-installation correction (Grok SHIP). Named API-key provenance, per-installation signature/allowlist verification, distinct principal namespaces and duplicate-assertion refusal are wired in the normal router.

Validation on this combined source: 5 OpenWebUI tests and 170 router tests passed on Spark, with forced rebuild and unchanged source manifests; formatting and CWE-532 scan (568 files, zero findings) passed. These groups overlap; counts are not additive. Live OpenWebUI/browser/provider journeys remain unproved. Existing unrelated CI/protocol/account-workflow blockers remain open.

Approved account design and test plan are now committed unchanged in 3794393a.

PR #539, opened 2026-09-13
chore(release): v4.0.0 integration line — CI visibility (draft, not for merge)

MikkoParkkola:
Draft, deliberately. This PR exists so the v4.0.0 integration line is testable at all; it is **not** a request to merge.

## Why it was needed

`.github/workflows/ci.yml` triggers on `push` only for `main` and `v*` tags, and on `pull_request` targeting `main`, `codex/v4-*`, or `fix/mrtr2-continuation-handle`. `fix/v4-integration-ci-green` is none of those, and its only PR (#531, into `codex/v4-next-integration`) merged at `97cb2856`. Every commit pushed to the line since then — including the two MIK-7387 commits `416a3319` and `93a77a80` — has never been seen by CI and could not be. A branch that cannot run CI reports neither green nor red; it reports nothing, and a release plan that lists "strict CI green" as a gate was pointing at a gate with no instrument behind it.

## What is in scope here

Only the signal. The merge decision for this line is sequenced in `docs/release/v4.0.0-gap-closure-plan.md` and comes after the merge queue (19 of 74 evidence commits are still off the line) and the scope re-cut — not here.

## What a green run does and does not establish

Green means the line compiles, lints and tests at this revision. It is not the DoD functional pass, which is still `NOT RUN` and requires the running gateway driven from the revision under review.


MikkoParkkola:
Triage recorded on #512: this branch carries 120 commits that #512 lacks and is missing exactly one, `47f2c23d` (#531, 20+ files). Cherry-picking that commit here makes #512 closable as superseded. Both branches still need a rebase onto `main` — `git merge-tree` reports 113 conflicting files against main's 15 new commits, which include the protocol-negotiation fix (#520), the Windows stdio environment fix (#522) and the mTLS signing-usage fix (#487). Detail: https://github.com/MikkoParkkola/mcp-gateway/pull/512#issuecomment-5657740690

PR #561, opened 2026-09-15
fix(meta-mcp): close the three losses the absent-declaration adjudication found

MikkoParkkola:
## Scope

`0f04a179` (squash of #473) added 1059 declarations to `src`; 32 are absent at the release tip, 13 of them production, and 10 of those had no trace at the tip by name. This adjudicates all ten against the tip's **behaviour** rather than its symbol table, at `docs/release/verify/0f04a179-absent-declarations-adjudication.md`, and then closes the three losses it found.

Result: **three moves, four replaced-by-design, three losses — all three losses now closed on this branch.**

`replaced-by-design` is a separate verdict on purpose — a client-visible delta is a release-note input and must not share a label with a relocation.

## The finding

An earlier draft called `is_mcp_envelope` a move on the strength of a matching body at `invoke.rs:216`. It had **two** call sites, and the survivor is the *older* one: the predicate inside `apply_validated_output` is pre-commit code the squash never touched, and deduplicating against it is why the helper was extracted at all. The site the commit added is gone; the line it replaced stands verbatim at `invoke.rs:160-161`.

The call-site rule that caught it: a helper extracted to dedup *n* sites is a partial loss unless all *n* survive. Matching on the body alone would have published this as safe.

## What the branch then fixed

**`is_mcp_envelope` restored (`MIK-7212.MRTR.2a`).** The reachability argument the adjudication left assembled and unrun was executed as one unit test, and it ran red: fed an interim envelope with nothing extractable, `enforce_output_schema` republished the backend's opaque `requestState` under `structuredContent` and inside the rewritten `content[0].text`, and the mint rewrites only the top-level field. The restored predicate requires `content` to be an array — the shape `extract_output_validation_target` actually consumes — and both call sites now route through it.

**The MRTR7 bridge restored (`MIK-7212.MRTR.7a`/`7b`).** The reconcile merge `b3f05fd4` resolved `invoke.rs` against the `main` side and dropped `BridgeDispatcher` and `TracingBridgeObserver`; `git log -S` hides it because the removal is a merge resolution. The hunks are re-applied with a case the earlier grade never had: `a_legacy_clients_question_is_bridged_from_the_invoke_path` drives a **registered backend and a real `ClientChannel`** through `invoke_tool`, rather than a `FakeBackend` through the bridge's own suite — the blindness that let 26 green cases sit over a bridge the gateway never entered. Deleting the branch turns the test red; restoring it goes green.

**Three defects found in the re-applied wiring, all fixed here.** A bridged completion committed the backend's own body to the idempotency key, where the path stores a withheld placeholder precisely because a gate below may block it — a gate rejection would have left blocked data under the key. A `BackendFailed` round returned without settling, releasing a key whose side effect may already have run; it now settles with `fail` and is pinned by `a_failed_bridged_round_settles_the_idempotency_key`. And every bridged round now passes `admit_spend`: the spend gate runs once per `gateway_invoke` while a bridged exchange makes one backend call per round, so a backend could otherwise walk past an operator's limit by asking.

One reported defect was rejected on evidence: `stopped_to_ask` does go stale after a bridged completion, but the resulting cache suppression is correct — the response-cache key covers `arguments`, not the answers the caller gave in-band, so caching a bridged body would serve the next identical call somebody else's reply. The reason is now written at the arm.

## Grades

- `MIK-7212.MRTR.2a` — **MET**, held by the restored guard rather than by the trace that first claimed it.
- `MIK-7212.MRTR.7a`/`7b` — **PARTIAL -> MET**, non-blocking, on the re-applied wiring plus the wired-path test and its negative control.

Burndown: **2 core open, 15 minor, 17 total, delta -2.** `NFR.SEC.7` and `NFR.PERF.1` are what remain.

## Follow-up: the reviewers' finding on the fix itself

Both reviewers came back on the bridge fix with the same defect, and it verified at source. A bridged round refused by the spend budget was reported as `BackendFailed` -- the one bridge error the settlement arm treats as a dispatch whose outcome is unknown, and therefore settles rather than releases. The refusal happens **above** `accounted_dispatch`, so nothing ran, and settling denied the caller a retry of work the backend never saw.

`d5674ca8` gives the pre-dispatch refusal its own variant, `NotAdmitted`, and leaves the settlement arm matching only on dispatched failures. Negative control run rather than reasoned: map the refusal back to `BackendFailed` and `a_budget_refused_bridged_round_releases_the_idempotency_key` fails on the cache-empty assertion; restore it and the suite is green.

The reviewers' second finding -- bridged-round cost warnings are discarded -- is real and out of scope. The drop is deliberate and documented at the call site; a warning channel for bridged rounds is a feature, not a release blocker.

## Gates

`cargo fmt --check` clean · `cargo clippy --all-targets -- -D warnings` zero · `cargo test --quiet` green · `cargo check --no-default-features` green.

## Review

Two independent reviewers (gpt, kimi) per revision. They converged on the `is_mcp_envelope` defect; the bridge pass produced the three idempotency and spend findings above. Every finding was verified at source before it was acted on, and the one that did not survive verification is recorded rather than silently dropped.


MikkoParkkola:
## Two-reviewer round on `3fa31ccd` — SHIP-WITH-FIXES from both, one blocker

Full write-up committed at `docs/release/2026-09-16-pr561-review-findings.md`.

**Blocker, confirmed at source.** `enforce_firewall_challenge`
(`src/gateway/meta_mcp/response_security.rs:176`) has no production caller —
every other occurrence in `src/` is a test module. The bridge construction site
at `src/gateway/meta_mcp/invoke.rs:2126` forwards the backend's elicitation,
sampling and roots requests to the client untouched, so a backend can put text
in front of a person that the configured firewall exists to refuse. This is not
a regression from the merge; it describes the bridge as it has always been on
this branch.

**Second item, both reviewers independently.** The `MIK-7212.MRTR.7a` and `.7b`
rows in the criteria ledger still say stdio is descoped to MIK-7387 and that
`tests/mik_7212_mrtr7_stdio_acs.rs` carries three `#[ignore]`s. It carries none,
and the concurrent-dispatch machinery is in `src/gateway/server/mod.rs` and
`src/gateway/server/stdio_channel.rs`. The ledger text is the defect, not the
code — but the rows have to be corrected or the package split out before the
release line can be graded honestly.

Three further findings are recorded as reported-not-verified: a stdout close that
leaves the stdio read loop admitting calls (`src/gateway/server/mod.rs:2395`),
the 65th dispatch permit blocking the sole stdin reader, and reply matching that
accepts an object with an `id` and neither `result` nor `error`.

**State otherwise green.** The merge of `origin/main` was clean,
`cargo test --test mik_7212_mrtr7_bridge_acs` is 28 passed / 0 failed, and CI on
this head is 44 passed, 4 skipped, 0 failed.

Reviewer availability: gpt and kimi ran. grok 1.0.30 refuses to start — its
sandbox cannot resolve the symlinked `/var/run/docker.sock`, and 1.0.30 is the
latest release. The glm endpoint answers 404 for its configured model.

MikkoParkkola:
## Why the tools/call test was failing

`ac_ext_1_d_a_declared_extension_is_recovered_on_the_tools_call_path` is byte-identical to its original at `c7555c8d` — nothing was dropped from the test. The regression is production-side: `c7555c8d`'s entire wiring was six lines in `handlers.rs`, and the criss-cross merge took the other side of that hunk. The `ExtensionSet`/`Extension` imports that went unused at `handlers.rs:37` were the *symptom* of the vanished call site, not dead code.

Restored at `handlers.rs:854`, reading main's `declared_extensions()` binding (already present at :848) rather than re-deriving it.

## File-size baseline: scope of the re-baseline

19 gate failures — 18 growths plus one new over-ceiling file. Every re-baselined row is a file this branch modifies, with two apparent exceptions that are not loosenings:

| File | Baseline | Now | Note |
|---|---|---|---|
| `src/backend/tests.rs` | 1914 | 1892 | **tightens** — main's row was a stale high-water mark |
| `src/gateway/meta_mcp/search_ranking_authz_tests.rs` | 1199 | 1199 | unchanged; appears in the diff only because `--update` reorders rows |

## Known debt, flagged not hidden

`tests/mik_7212_mrtr7_stdio_acs.rs` is under the ceiling on `main` and reaches **1272 lines** here — the one genuinely new over-ceiling file, now recorded rather than silently admitted. It wants splitting; that is not done in this PR because the cargo disk guard is active locally, and an unverifiable split of a 1272-line integration test pushed straight to CI is a worse trade than recording the debt.

MikkoParkkola:
## Merge-resolution review record

The criss-cross merge produced one semantic collision: what a duplicate retry is *served* after a bridged round fails post-dispatch. Both sides prevent re-execution; they differ only in representation.

- Branch (`ce3a8e9d`): `reservation.fail(...)` — the retry receives a `-32003` error.
- Target (`c40ed24a`, #571): `commit(&withheld_side_effect())` — the retry receives a marker body.

**The target's behaviour is kept.** This is not a new design decision: `c40ed24a` is already on `main` with its own guard (`a_failed_bridged_retry_does_not_free_the_idempotency_key`). Preserving it is the default outcome of a merge; overriding it would have been the change needing justification.

### Verified at source

| Item | Evidence |
|---|---|
| Union classifier is total | `invoke.rs:980-994` — `is_pre_dispatch()` diverts every provably-unexecuted case to `NotAdmitted`; the `else` arm is unconditionally `MayHaveActed` |
| `Dispatch::NeverReached` has no producers | Superseded by `NotAdmitted`, which carries a message and releases the key; rationale recorded at `invoke.rs:987` |
| Assertion retarget at `tests.rs:6401` still pins non-readmission | Mutation check below |

### Mutation check on the retargeted assertion

The concern raised in review was that changing `expect_err` → `expect` + marker assertion trades a real guard for a string match. Removing the marker commit at all three sites (`invoke.rs:2137`, `:2232`, `:2338`) turns both tests red, for the correct reason:

```
a_failed_bridged_round_settles_the_idempotency_key
  panicked: backend called more times than the script has answers

a_failed_bridged_retry_does_not_free_the_idempotency_key
  assertion `left == right` failed: the key was freed after a round that may
  have acted, so the duplicate re-executed
  left: 3, right: 2
```

The assertion still discriminates re-execution, not text. Restored; both pass.

### Coverage limits

Two external review seats returned no verdict for infrastructure reasons — one produced no output, one was unauthenticated. One seat returned findings; the two that concern the success-shaped replay apply to `main`'s existing contract rather than to this merge, and are filed separately.

### Follow-up, not blocking

`Dispatch::NeverReached` is now unreachable and should be deleted once this lands.

Issue #527, opened 2026-09-11
One parser for a configured stdio command, used by spawn and doctor alike


MikkoParkkola:
Fixes #523.

## Source

Five sites derive a program name from the same configured `command` string, with two different parsers:

| site | parser |
| --- | --- |
| `src/transport/stdio.rs:146` (spawn) | `shlex::split` |
| `src/backend/lifecycle.rs:61` (runtime-profile hint) | `shlex::split` |
| `src/security/http_diagnostics.rs:125` | `shlex::split` |
| `src/gateway/ui/backend_ops.rs:389` | `shlex::split` |
| `src/commands/doctor.rs:459` (doctor) | `split_whitespace` |

## Impact

Two defects, one cause.

`shlex::split` is POSIX, so on Windows a backslash escapes rather than separates: `C:\Windows\py.exe -3.11 -m markitdown_mcp` becomes `C:Windowspy.exe`, and the spawn fails with "program not found". Forward slashes work, which is why the configuration that every Windows tool writes is the one that breaks.

`doctor` parses differently from the code that will run the command, so it reports `'C:\Windows\py.exe' found` for a command that cannot spawn. The same divergence is wrong in the other direction on every platform: a quoted path containing spaces (`"C:\Program Files\py.exe"`, or a macOS path under `/Applications/My App/`) makes `doctor` look up a truncated binary and report a false failure. A preflight check that parses differently from the thing it checks cannot be trusted on any input.

## Acceptance

- One `split_command()` helper is the only parser of a configured `command` string; all five sites call it.
- On unix it is `shlex::split` and existing behaviour is unchanged, proven by the current tests staying green untouched.
- On Windows it follows `CommandLineToArgvW`: double quotes group, and backslashes act as escapes only in a run immediately preceding a double quote. `C:\Windows\py.exe` and `"C:\Program Files\py.exe"` both survive.
- Fixtures for the reported string, a quoted path with spaces on both platforms, and an unterminated quote (which must stay an error, not a silent truncation).
- `doctor`'s stdio check reports on the same program name the spawn will use; a config that cannot spawn cannot produce a PASS.
- The configuration documentation states that `command` is parsed with the host platform's rules.

## Fail-fast

[...]

PR #564, opened 2026-09-16
fix(stdio): one host-platform split_command for spawn and doctor

### LINKED ISSUE #523: Windows: stdio backend command paths with backslashes break under POSIX shlex; doctor false-PASS
## Problem
Stdio backend `command` strings are parsed with POSIX `shlex::split` (`src/transport/stdio.rs`). On Windows, unquoted paths with backslashes are treated as escapes:

- Config: `C:\Windows\py.exe -3.11 -m markitdown_mcp`
- After shlex: `C:Windowspy.exe` → `Failed to spawn: program not found`

Forward slashes (`C:/Windows/py.exe`) work.

## Doctor false positive
`doctor`’s stdio check uses `command.split_whitespace().next()` (not shlex), so it reports `'C:\Windows\py.exe' found` even though runtime spawn will fail after shlex.

## Expected
Either:
1. Document clearly that Windows `command` paths must use `/` or be quoted for shlex, and make `doctor` use the same parser as spawn; or
2. On Windows, parse commands in a Windows-aware way (or normalize `\` in the program path before spawn).

## Environment
- OS: Windows 10/11
- Gateway used as Cursor MCP via `serve --stdio`

Happy to send a small doctor-alignment PR if that direction is preferred.

yfcyfc123234:
## Summary
Implements #527 / fixes #523.

- Add `transport::split_command()` as the only parser of a configured stdio `command` string.
- Unix: `shlex::split` (unchanged behaviour).
- Windows: `CommandLineToArgvW` rules so `C:\Windows\py.exe …` keeps backslashes; quoted paths with spaces survive.
- Both rule sets are pure functions and unit-tested on every host (`split_command_unix` / `split_command_windows`); only the `cfg` selection is host-specific.
- Wire the five sites: stdio spawn, lifecycle executable hint, `summarize_stdio_command`, UI `summarize_command`, and `doctor` stdio check.
- Doctor now fails on invalid quoting / empty command instead of a false PASS from `split_whitespace`.
- Document host-platform parsing in README / example configs.

## Regression fixture (from our report)
```
C:\Windows\py.exe -3.11 -m markitdown_mcp
```

## Test plan
- [x] `cargo test --lib transport::command_split`
- [x] `cargo check --bins`
- [ ] CI on this fork PR (maintainer may need to approve workflows)


yfcyfc123234:
﻿Pushed a follow-up commit to clear CI on this PR:

- cargo fmt for doctor.rs / command_split.rs
- bump rustls to 0.23.45 for RUSTSEC-2026-0285 (Dependency audit + Security gate)

Tests / Clippy / Windows check were already green.

MikkoParkkola:
Merged — thank you. This was a clean, well-reasoned contribution and the review found nothing to push back on.

Specifically worth calling out:

- The `CommandLineToArgvW` rules are implemented correctly, including the `2n` / `2n+1` backslash-before-quote cases and the doubled-quote extension. An unterminated quote returning `None` rather than a truncated argv is the right call.
- Keeping both rule sets as pure functions tested on every host, with only `split_command` behind `cfg`, means Linux CI actually exercises the Windows path. That is what makes this maintainable rather than a platform blind spot.
- Collapsing all five call sites onto one parser removes the drift that caused #523 in the first place, and `doctor` failing on invalid quoting instead of reporting a false PASS is a real improvement beyond the reported bug.
- The rustls 0.23.45 bump for RUSTSEC-2026-0285 is welcome on its own merits; the repository needed it regardless.

Shipping in 4.0.0. #523 closes with this.

PR #566, opened 2026-09-17
[THROWAWAY PROBE — do not merge] MIK-7387 stdio bridge census

MikkoParkkola:
Throwaway branch. Two `warn!` lines and nothing else, opened as a draft only because CI runs on `pull_request` and not on a branch push.

Purpose: the `NoSession` match in `src/gateway/meta_mcp/invoke.rs` is silent by construction, so a hit on it is invisible in any CI log. Two red rows (`ac_mrtr_7a_*`, `ac_mrtr_7b_*`) pass 6/6 locally and fail only in CI, so the mechanism can only be read in the environment that fails.

This PR will be closed and the branch retired once the run is read. It must not be merged.


MikkoParkkola:
Closing: the probe was inert. Both `tracing::warn!` calls in ccd7669f emitted nothing in CI because no test installs a subscriber and the workflow sets no RUST_LOG, so zero hits is no evidence either way. The failing test's own census already histograms error codes and messages, which is a working instrument, so the eventual fix carries its own measurement and needs no probe branch. Findings recorded in docs/requirements/RELEASE-4.0.0-readiness-plan.md (3a7d2865).

PR #499, opened 2026-09-08
feat(v4): compose durable tasks with signing and shared admission

MikkoParkkola:
Durable tasks and synchronous tool calls now share one admission authority, so the same owner and idempotency key cannot dispatch once through each mode. Continuations retain the original client key while partitioning individual input rounds. This composition also preserves signed task responses, authenticated destructive confirmation, native task result shapes, and the approved shared owner for auth-disabled tasks.

This is a draft integration checkpoint combining the task and signing branches. It targets `fix/mrtr2-continuation-handle`. Related foundation PR #495 remains open while the canonical integration is reconciled.

Validation on commit `35ab20dafa5ed1813e09e3dd22e40fbbbd3b1289`:

- `cargo check --all-targets --all-features --jobs 1` passed.
- 45 route tests passed, including seven joint signing/task cases with independent JCS/HMAC verification.
- Three startup tests passed, including restoring durable task bindings into a fresh shared admission authority.
- All 20 TASK.1 acceptance tests passed, including paired auth-enabled refusal and auth-disabled execution.
- Source hashes remained unchanged throughout each run; no resource failures occurred.

Remaining release gates include strict warning cleanup, integrated coverage/mutation qualification, the queued-handoff drain race, the worker policy recheck barrier, and the remaining task notification, recovery, expiry, and upstream lifecycle increments. These 68 passing tests do not establish full task lifecycle or release acceptance.

PR #568, opened 2026-09-18
fix(docker): the published image exits on startup, and CI never ran it

### LINKED ISSUE #529: Close the push-before-sign window on both ghcr publishers
## Source

- `.github/workflows/docker.yml`, job `build`: `Build and push` runs before `Cosign keyless-sign the pushed image by digest`.
- `.github/workflows/ci.yml`, job `docker`: same order.
- Disposition already recorded in `docs/release/v4.0.0-prerelease-channel.md`, section "What this does not close".

## Problem

Both workflows fire on a `v*` tag, both push `ghcr.io/mikkoparkkola/mcp-gateway:VERSION` from the same commit, and neither can depend on the other. Each signs and SBOM-attests the digest it pushed, so whichever push lands last carries a signature — unless the run fails or is cancelled between the push and the signing steps. In that window the tag is public and unsigned.

The run goes red and a consumer running `cosign verify` gets a non-zero exit, so the state is detectable rather than silent. It is not prevented.

## Acceptance criteria

- [ ] On a tag ref, neither publisher makes a tag name resolvable until the digest behind it has a cosign signature and an SBOM attestation.
- [ ] A forced failure injected between push and signing leaves no `:VERSION` tag in the registry.
- [ ] Branch and PR builds keep their current tagging behaviour (`:main`, `:SHA`) and stay unsigned.
- [ ] `scripts/release/test_check_tag_manifest.py` gains a case that fails against the current ordering and passes against the fix.

## Fail-fast check

`docker buildx imagetools create --prefer-index=false -t NAME:VERSION NAME@DIGEST` must resolve `:VERSION` to the *same* digest that was signed. The default `--prefer-index=true` does not carbon-copy a single source (`buildx` `commands/imagetools/create.go`), so a promotion that synthesises a new index would sign one digest and publish another — strictly worse than today. Verify this against a scratch registry before touching either workflow; if the digest is not preserved, close this as won't-fix and keep the documented window.

## Impact

Supply-chain: a released tag that verification rejects. Frequency is bounded by how often a release run dies in a two-step window, so the expected loss is small; the cost of the bad outcome is a release users cannot verify, and the fix is a contained change to two workflows.

### LINKED ISSUE #536: Release image publisher in ci.yml bypasses the trivy CVE gate
## Source

`.github/workflows/ci.yml:328` — job `docker`, `if: startsWith(github.ref, 'refs/tags/v')`, build-and-push at `.github/workflows/ci.yml:364`.

## Evidence

The image CVE gate lives only in `.github/workflows/docker.yml`: build to a local `:scan` tag, run `aquasecurity/trivy-action` with `exit-code: 1` on HIGH/CRITICAL, then push. `ci.yml`'s `docker` job has no equivalent. Its steps go Extract tag → Log in to GHCR → Build and push → cosign → syft, and its `needs:` list covers `audit` (Rust dependency scan) but no image scan. So on a release tag it pushes `ghcr.io/mikkoparkkola/mcp-gateway:$VERSION` and, for a stable release, `:latest` without any base-image CVE check.

The two jobs also race. A comment already in `docker.yml` records it: on a tag both workflows push the same name from the same commit, neither can depend on the other, and the last push owns the name. A `docker.yml` build that trivy rejected does not stop `ci.yml` from publishing its own image under the same tag.

## Impact

The tags users actually pull — `:latest` and the version tag — are the ones with no image CVE gate in front of them. The gate covers pull requests and branch pushes, which is where it matters least. Everything downstream is signed and attested against that digest, so cosign and the SBOM certify an image nothing scanned.

## Acceptance

- The `docker` job in `ci.yml` builds the image, runs the same trivy configuration as `docker.yml` (`exit-code: 1`, `severity: HIGH,CRITICAL`, `ignore-unfixed: true`, `trivyignores: .trivyignore`), and pushes only if that step passes.
- The scanned image is the pushed image: one build, shared cache, matching build args.
- No release tag can publish an image that failed a scan, from either workflow.

## Fail-fast check

Plant a fixable HIGH CVE (pin an old `ca-certificates`, or point `trivyignores` at an empty file while a known-fixable finding is present) and push a `v*` tag to a scratch repo. Today the image publishes. After the fix the job fails before the push and no tag moves.

## ROI

Cost: one trivy step plus a build reorder in one job. Value: closes the gap between what the gate inspects and what users install. Without it the CVE gate is a pull-request formality — it can never block the artifact it exists to block.

## Origin

Found while fixing a related defect in the same gate (#535): the runtime stage's `apt-get upgrade` was being restored from the build cache, so the scan reported CVEs the upgrade already fixed. That fix does not touch this bypass; keeping them separate keeps each reviewable.

MikkoParkkola:
## What

The published container image for the v4.0.0 release commit exits 1 on startup, and no job in this repository has ever started a container, so it shipped green.

**The defect.** `ghcr.io/mikkoparkkola/mcp-gateway:a2505be` fails with:

```
WARN  task store directory not created error=Permission denied (os error 13) path=/home/gateway/.mcp-gateway/tasks
ERROR Gateway error: Configuration error: task store at '/home/gateway/.mcp-gateway/tasks' could not be opened: task store unavailable
```

`useradd -r` creates a system account with no home directory. Three features resolve paths under `$HOME` — the task store, the skill registry (`src/skills/registry.rs:73`) and the chain checkpoint store (`src/chains/checkpoint.rs:76`) — and the task store treats an uncreatable parent as fatal. It fails before configuration is read, so no operator config avoids it. Fixed at source with `-m -d /home/gateway`, which covers all three rather than patching the store.

**Why every gate missed it.** `docker.yml` builds, Trivy-scans, pushes, cosign-signs by digest and syft-attests an SBOM. All five inspect the image at rest. A green pipeline established that the image compiles, packages, scans clean and is signed — and nothing about whether it runs.

## Evidence

The new smoke step was verified two-sided against the real artifact:

| image | gate |
|---|---|
| `ghcr.io/mikkoparkkola/mcp-gateway:a2505be`, unchanged | exit 1, `image exited on startup` |
| that same image plus `/home/gateway` and nothing else | exit 0, HEALTHCHECK reports `healthy` |

The second image is the published one with only the directory `useradd -m` creates, layered on top — so the passing side tests this diff's mechanism, not a rebuild.

The gate reads the container's own `HEALTHCHECK` verdict rather than re-probing `/health`, so a HEALTHCHECK aimed at the wrong port cannot stay green. `--rm` is deliberately absent: it would delete the container before the failure branches read its logs.

## Out of scope

The image is `linux/amd64` only (`platforms` appears nowhere in `docker.yml`), so it cannot run natively on Apple Silicon. The smoke step runs on the runner's own architecture and cannot observe this. Multi-arch publishing needs a build matrix on a native `ubuntu-24.04-arm` runner with a manifest merge, restructuring a job that also gates the tag, signs by digest and attests an SBOM — a pipeline change, not a line. Tracked on MIK-7482; `NFR.PKG.1` stays blocking until it lands.

## Review

kimi-review: SHIP. Its cleanup finding is taken as a `trap` (its suggested `--rm` would have blinded the failure diagnostic); its HEALTHCHECK-lockstep improvement is taken. gpt-review and grok-review produced no verdict this run — quota exhaustion and a watchdog truncation respectively, both wrapper failures rather than review outcomes.

---

## Issues closed

`Closes #529` — the push-before-sign window. Both acceptance clauses that this repository can still satisfy are met: the release names are created only from an index whose digest already carries a cosign signature and an SBOM attestation, and a failure injected anywhere between the build and `cosign verify` leaves no `:VERSION` in the registry, because no release name exists until after verify. The issue names two publishers; there is now one. `docker.yml` no longer pushes on a `refs/tags/v*` ref, which its own regression test locks, and its branch and PR tagging (`:main`, `:SHA`, unsigned) is unchanged — the third clause. On the issue's fail-fast check: it predicted that the default `--prefer-index=true` would synthesise a new index rather than carbon-copy a single source, and asked for a scratch-registry check before touching either workflow. Checked on Docker 29.5.2 / buildx v0.37.0 against a throwaway `registry:2`; the digest was preserved on all three tags without `--prefer-index=false`, so the flag is not added. CI does not rely on that holding: the final step re-inspects `:${VERSION}` and fails the job if it resolves to anything but the signed digest.

`Closes #536` — the release publisher bypassing the trivy CVE gate. Verified in the YAML rather than taken from the comment that claimed it, and the claim holds. `docker-build` builds push-by-digest with no tag, pulls that digest back, and runs `aquasecurity/trivy-action` against `ghcr.io/mikkoparkkola/mcp-gateway@${{ steps.build.outputs.digest }}` with `exit-code: "1"`, `severity: HIGH,CRITICAL`, `ignore-unfixed: true`, `trivyignores: .trivyignore` — the configuration the issue asks for, matching `docker.yml`'s. It runs on both matrix legs (`amd64` on `ubuntu-latest`, `arm64` on `ubuntu-24.04-arm`) with `fail-fast: true`, so a CVE on either architecture stops the job before `docker-manifest` runs and therefore before any name exists. The scanned image is the pushed image by construction, not by cache luck. The issue's third clause — "no release tag can publish an image that failed a scan, from either workflow" — holds for the same reason #529 does: there is one publisher now.

**#570 stays open.** It names three one-sided guards; this branch fixes one of them. Item 2, attribution by substring, is gone: `steps()` now parses per job and takes a `job` filter, so two jobs carrying an identical step body attribute to their own job — but the fixture the AC asks for is not written. Items 1 (a `check_tag_manifest.py` invocation added to `docker-build` must be CAUGHT) and 3 (a mutation binding `LIST` to `steps.*.outputs.amd64` must be CAUGHT) are untouched.

## Second review round

Four further findings, each with a test that was red before its fix:

- **Supply chain.** `publish-mcp-registry` fetched `mcp-publisher` from `releases/latest` through a pipe into `tar`, then gave it an OIDC token with publish rights to the public MCP Registry under this project's name. Pinned to `v1.8.1` and verified against a per-architecture SHA-256 before first execution; `curl -fsSL` to a file rather than a pipe. A pin alone would not close it — release assets can be replaced in place — so the checksum is the control, and a version bump now has to move the digest in the same commit.
- **The smoke gate proved half of NFR.PKG.1.** The row reads "the container image the release publishes starts, and serves an MCP request from outside the container"; the gate read only the image's own HEALTHCHECK, which probes from inside, where a loopback bind answers and a published port reaches nothing — the exact defect the row records. A second leg publishes the port and sends a real `initialize` from the runner. The assertion is a JSON-RPC envelope, not a 200: a protocol-version mismatch answers with an `error` member and still proves the image served an MCP request from outside, while a refused connection, a hang or an empty body do not.
- **The default invocation was never exercised.** Every leg injected a synthetic `--config`. The image ships no `/config.yaml`, and the HTTP serve path is fail-loud on an explicit config it cannot load, so the intended no-config behaviour is a refusal that names the missing file — inventing a default config would start a gateway with a posture nobody chose. The third leg asserts non-zero exit, `/config.yaml` in the output, and no panic.
- **A degenerate HEALTHCHECK would not have been caught.** The gate's whole verdict is the image's own probe, so `CMD true`, `NONE`, or one aimed at a port nothing serves would keep it green. The probe is now read from `.Config.Healthcheck.Test` and required to name `/health` and port 39400 before any container starts.

Two of the new guards were themselves one-sided when first written — a `docker run` spans continuation lines, and a line-anchored match read only the first, so an assertion about a later argument passed whether or not the argument was there. Both were rewritten to fold continuations before matching, and re-confirmed red.

## Verified here, not in CI

The container legs cannot run in this environment: they need a built image, and `cargo` was out of scope for this change. `bash -n` and `shellcheck` are clean on the gate, and the wiring assertions are green, but the three legs have never executed against a real image. `docker-manifest` and `publish-mcp-registry` only run on a `v*` tag, so the pinned download, the copy-by-digest publish and the smoke legs all first execute on a real release.

## Merge order

The criteria-ledger rows are isolated in the tip commit, `docs(release): the criteria-ledger rows for this branch, in one commit`. #499, #561 and #568 have that one file in common and nothing else, so the three ledger commits can be consolidated once with the tally recounted once, rather than each merge re-conflicting the other two. Rebased onto `main` after #578: clusters M and N are main's and are not resurrected, cluster O is this branch's, and the headline is the script's recount — 149 criteria, 190 rows, 188 met or non-blocking, 2 blocking.


MikkoParkkola:
### Second publisher closed

The first version of this branch gated only `docker.yml`. An adversarial review of it surfaced the hole: `.github/workflows/ci.yml` has its own `docker` job that builds, pushes, cosign-signs and SBOM-attests `ghcr.io/mikkoparkkola/mcp-gateway:<version>` on a release tag. It fires on the same tag independently of `docker.yml` and cannot be blocked by it, so a gate in one workflow says nothing about what the other publishes.

Both publishers now call the same gate, `scripts/ci/smoke-image.sh`:

| publisher | fires on | gate position |
|---|---|---|
| `.github/workflows/docker.yml` | push, PR, tag | after the local load and Trivy scan, before any push |
| `.github/workflows/ci.yml` (`docker` job) | release tag | after a `load: true` build, before the push-and-sign build |

`ci.yml` needs the extra local build because a container can only be started from a loaded image; it shares `cache-from`/`cache-to: type=gha,mode=max` and the identical `APT_CACHE_BUST` build-arg with the push build below it, so the push is a cache hit rather than a second compile.

The gate reads the container's own `HEALTHCHECK` verdict rather than re-probing `/health` itself, so the line the image ships is the line under test — a `HEALTHCHECK` aimed at the wrong port would otherwise stay green forever. An image declaring no `HEALTHCHECK` at all is refused in seconds instead of timing out at 90 with a misleading message, and an image that never loaded is reported as that rather than as a missing `HEALTHCHECK`.

Verified locally, four ways:

| input | expected | result |
|---|---|---|
| published `:a2505be` (no `$HOME`) | fail | exit 1, `::error::... exited on startup` |
| same image built with this branch's `useradd -m` | pass | exit 0, `... its own HEALTHCHECK reports healthy` |
| image with no `HEALTHCHECK` | fail fast | exit 1, `declares no HEALTHCHECK` |
| image never loaded | fail fast | exit 1, `cannot inspect ...` |

Still out of scope and still blocking `NFR.PKG.1`: the image is `linux/amd64` only. That needs a native-arm64 build matrix and a manifest merge across a job that also gates the tag and signs by digest — a reviewed design, not a line appended here. MIK-7482 carries it.

MikkoParkkola:
## Both confirmed review findings fixed

### 1. The push-before-sign window was open, and the code claimed it was closed

`docker.yml:184` said "the tag is created only after both architectures are signed". It was not: `docker-manifest`'s *Create the manifest list* step ran `imagetools create --tag …:${VERSION} [--tag :latest]` **before** *Install cosign*, so the public release name was pullable and unsigned for the whole sign + SBOM + verify span. `cosign verify` passed afterwards regardless — it reads digests, never the tag.

Fixed by keeping every signature keyed to the digest and moving only the *naming*:

1. The index is created under a provenance-only tag, `:sha-${GITHUB_SHA}`. No release name exists yet.
2. The both-platforms assertion now runs against `${IMAGE}@${LIST}`, not `:${VERSION}`.
3. cosign sign, SBOM attest and cosign verify run on the digests, unchanged.
4. **Only after verify passes**, the release names are put on the already-verified index by digest: `imagetools create --tag :${VERSION} [--tag :latest --tag :MAJOR.MINOR] "${IMAGE}@${LIST}"`.
5. A final step asserts `imagetools inspect :${VERSION} --format '{{.Manifest.Digest}}'` equals `${LIST}` — the release tag provably resolves to the signed digest.

Visible side effect: `:sha-<commit>` is now a permanent public tag in GHCR. It is deliberate — it is the handle the signing steps need before a release name exists, and it gives every release a provenance name that never moves.

The registry-name annotation stays on the **first** create only. Annotating rewrites the index and changes its digest, so repeating it on the copy would publish bytes no signature covers.

### 2. The stable `:MAJOR.MINOR` tag was dropped

`main`'s `docker.yml` published `type=semver,pattern={{major}}.{{minor}}`; this PR disabled that publisher and the replacement published only `:${VERSION}` and `:latest`, so consumers pinned to `:4.0` would have stopped receiving releases silently. It is published again, derived from `${VERSION}` with `cut -d. -f1,2`, inside the *same* `if [ -n "${LATEST_TAG}" ]` branch as `:latest` — one guard, so a prerelease cannot move one pointer without the other. `cut` rather than `${VERSION%.*}` so the derivation stays correct on a suffixed version if it is ever moved out from behind that guard; today it cannot be reached by a prerelease at all.

### Tests — written failing first

In `scripts/release/test_check_tag_manifest.py`:

- `test_the_release_tag_is_created_only_after_the_signature_verifies` — step **order by index** inside the `docker-manifest` job: every release-tag create sits after the last `cosign verify`; the first create names only the `sha-` provenance tag and contains none of `${VERSION}`, `latest`, `LATEST_TAG`, `MAJOR_MINOR`; and a later step compares the published digest to `${LIST}`.
- `test_the_stable_channel_keeps_its_major_minor_pointer` — exactly one major.minor value derived from `${VERSION}`, derived and used inside `:latest`'s own guard, and used nowhere outside it.
- `test_the_branch_builder_still_refuses_to_push_on_a_tag` — regression lock on `docker.yml`'s `push:` condition. **This one is green today**, by design: it locks an invariant the old test only inferred from the *absence* of cosign commands.

`steps()` gained a `job` filter, because order across two jobs is not an ordering Actions honours.

Evidence they bite, against the pre-fix `ci.yml`:

```
FAIL: test_the_release_tag_is_created_only_after_the_signature_verifies
AssertionError: [] is not true : ci.yml: the only imagetools create is the one that publishes a tag

FAIL: test_the_stable_channel_keeps_its_major_minor_pointer
AssertionError: Regex didn't match: ... : ci.yml: the major.minor tag is not derived from ${VERSION}

Ran 42 tests — FAILED (failures=2)
```

Six new cases in `test_workflow_wiring_mutations.py`, including one that exercises the ordering assertion *itself* rather than a string check — it copies the release tags in an earlier step and leaves the late copy in place, so every string the suite looks for is still where it was:

```
AssertionError: 6 not greater than 11 : ci.yml: - name: Assert the list carries both platforms publishes a release tag before cosign verify
```

### Gates

```
python3 scripts/release/test_check_tag_manifest.py        Ran 42 tests — OK            exit 0
python3 scripts/release/test_workflow_wiring_mutations.py 87 cases agree              exit 0
actionlint 1.7.12 .github/workflows/{ci,docker}.yml       (no output)                 exit 0
python3 scripts/release/count-release-criteria.py --check 149 criteria, 190 rows      exit 0
```

`cargo` was not run (out of scope for this change; no Rust touched).

### Copy-by-digest, checked rather than assumed

The publish step rests on one property: `imagetools create` with a single index source and no annotations copies the descriptor instead of composing a new index, so the digest survives. Checked locally against a throwaway `registry:2` (Docker 29.5.2, buildx v0.37.0): two single-platform digests → annotated index under a `sha-` provenance tag → that index copied by digest onto `:4.0.0`, `:latest`, `:4.0`; all three resolve to the source digest `sha256:0862671115cf…`. CI does not take it on trust either — step 5 asserts it, so a future buildx that rewrites on copy fails the release job rather than publishing a tag no signature covers.

### Beyond the two findings

`docs/release/v4.0.0-prerelease-channel.md` still described two ghcr publishers and told consumers to try `ci.yml` **and** `docker.yml` signer identities in a loop. This PR made `ci.yml` the only publisher, so that recipe was already stale; it now names one identity and the table lists `:4.0` again. Note this does not verify images an earlier `docker.yml` signed — pre-4.0.0 digests still carry that identity.

Prose updated to match the code: `docker.yml`'s comment block, the metadata-action rows note, `docs/design/2026-09-18-multi-arch-container-publishing.md` (mechanism, `#529` closure, acceptance 6 and 7, falsifier).

MikkoParkkola:
## Second round: four findings, and the issue verdicts

Kimi's review found four more one-sided guards. Each is fixed with a test that was red against the pre-fix tree — full detail is in the PR description; the short version:

| # | Finding | Fix |
|---|---|---|
| 1 | `publish-mcp-registry` piped `mcp-publisher` from `releases/latest` straight into `tar`, then handed it an OIDC token with publish rights to the public registry | pinned to `v1.8.1`, SHA-256 verified before first execution, `curl` to a file rather than a pipe |
| 2 | The smoke gate read only the image's own HEALTHCHECK, which probes from inside — where a loopback bind answers and a published port reaches nothing, the exact defect `NFR.PKG.1` records | second leg publishes the port and sends a real `initialize` from the runner |
| 3 | Every leg injected a synthetic `--config`; the default invocation an operator actually types was never run | third leg asserts non-zero exit, `/config.yaml` named in the output, no panic |
| 4 | A degenerate HEALTHCHECK (`CMD true`, `NONE`, wrong port) would have kept leg 1 green forever | the probe is read from `.Config.Healthcheck.Test` and must name `/health` and 39400 before any container starts |

Two of those new guards were themselves one-sided when first written — a `docker run` spans continuation lines and a line-anchored match read only the first — so they were rewritten to fold continuations, and re-confirmed red.

**Issues.** `Closes #529` and `Closes #536` are now on the description, both verified at source rather than taken from a comment: #536's trivy gate scans the pushed digest on both matrix legs with `exit-code: "1"` and `fail-fast: true`, so a HIGH/CRITICAL on either architecture stops the job before any name exists. #529's remaining clause — one publisher, names created only after `cosign verify` — is what this PR does.

**#570 stays open**, with one of its three items fixed. Item 2 (attribution by substring) is gone: `steps()` parses per job and takes a `job` filter. Items 1 and 3 are untouched, and item 2's own fixture test is not written.

**Not executed here.** The three container legs need a built image; `cargo` was out of scope. `shellcheck` and `bash -n` are clean, the wiring assertions are green, but the legs themselves — and the pinned download, and the copy-by-digest publish — first run for real on a `v*` tag.

```
python3 scripts/release/test_check_tag_manifest.py         Ran 46 tests — OK             exit 0
python3 scripts/release/test_workflow_wiring_mutations.py  90 cases agree                exit 0
actionlint 1.7.12 .github/workflows/{ci,docker}.yml        (no output)                   exit 0
shellcheck scripts/ci/smoke-image.sh                       (no output)                   exit 0
python3 scripts/release/count-release-criteria.py --check   149 criteria, 190 rows        exit 0
```

Issue #569, opened 2026-09-18
MRTR.7 bridge: pin the stdio carve-out, the NoSession fallthrough, and make RoundsExhausted resumable


MikkoParkkola:
The MRTR.7 input bridge is back on the production invoke path (`a1c12b2f`), and both criteria rows are graded MET. An independent review of that change raised three follow-ups that are not defects at the current gate but leave real behaviour unpinned.

## Source

- `src/gateway/meta_mcp/invoke.rs:2007` — bridge call site inside `invoke_tool_traced`
- `src/gateway/meta_mcp/invoke.rs:801` — `BridgeDispatcher`, the production `BackendInvoker`
- Review: independent static review of `5c73a91c..7d235c61`, three IMPROVEMENT entries, verdict SHIP

## 1. Pin the two-halves invariant that keeps stdio out of the bridge (MIK-7212.WIRE.10)

stdio transports stay out of the bridge because of two independent conditions: the
non-empty-requests guard, and `plan` being positioned before `ask`. Either one can be
edited alone, and the descoped path silently changes behaviour instead of failing a test.

**AC:** a test asserts a stdio caller with a non-empty pending-requests interim takes the
`NoSession` arm; removing either half of the invariant turns it red.

## 2. Regression test for the `NoSession` fallthrough

The fallthrough is required to preserve the pre-bridge behaviour — continuation minted,
idempotency reservation declined. Today that is established by reading the code, not by a
test.

**AC:** a test drives an interim through the `NoSession` arm and asserts both the
continuation is minted and the reservation is declined.

## 3. Resumable continuation on `RoundsExhausted`

A bridge failure after progress returns a bare `-32003`, stranding the backend's
accumulated `requestState`. Minting a continuation that carries it makes the exchange
resumable. Requires the progress signal out of `run`, currently marked `ponytail:`.

**AC:** a multi-round exchange that exhausts its round budget returns a continuation
carrying the backend `requestState`, and resuming it reaches the backend with that state.

## Fail-fast

Each item is a failing test first: write the test against current `main`, watch it go red,
then make it green. Items 1 and 2 are test-only. Item 3 changes the `Err(error)` arm and
must not alter the `NoSession` path (item 2's test is the guard).

## ROI

Items 1–2: SMALL cost, they convert two silent failure modes into build failures — the
same class of miss that cost this feature its production wiring in a reconcile merge.
[...]

PR #571, opened 2026-09-18
fix(mrtr): put the MRTR.7 input bridge back on the production invoke path

MikkoParkkola:
The MRTR.7 input bridge had been wired into `invoke_tool_traced`, then a reconcile merge dropped the wiring. The AC suite `tests/mik_7212_mrtr7_bridge_acs.rs` constructs `InputBridge` itself, so all 28 cases stayed green over a bridge the gateway never entered — the suite covered the component, not the path.

## What changed

- `BridgeDispatcher` (`src/gateway/meta_mcp/invoke.rs:801`) and `TracingBridgeObserver` (`:874`) restored as the production `BackendInvoker`/observer, with the call site back inside `invoke_tool_traced`.
- A bridged round settles its idempotency reservation with the withheld-side-effect marker instead of the completed body, so a retry of the same key is told the side effect ran rather than being served a success a later gate withheld.
- `BridgeError::BackendFailed` — the one bridge error that fails past dispatch — now settles the reservation too. It previously left it on `Release`, so `Drop` readmitted the key and a duplicate submission re-executed the side effect. `Delivery`, `NoSession` and `Refused` fail before another dispatch and keep their release-on-drop default.
- `MIK-7212.MRTR.7a` and `7b` regraded MET and non-blocking; cluster M closed in the blocking rollup.

## Evidence

Both new tests enter through the production entry point and were red before their fix:

- `legacy_caller_is_asked_through_the_bridge` — red with `JsonRpc { code: -32003, … this exchange cannot be continued for this caller }`, the continuation mint answering in place of the bridge.
- `a_failed_bridged_retry_does_not_free_the_idempotency_key` — red at 3 backend dispatches against an expected 2.

`cargo test --lib gateway::meta_mcp` 656 passed / 0 failed · `cargo test --test mik_7212_mrtr7_bridge_acs` 28 passed / 0 failed · `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean · `scripts/release/count-release-criteria.py --check` reports 187 met or non-blocking, 2 blocking.

## Out of scope

stdio transport bridging (MIK-7387). Three follow-ups tracked in #569: pin the two-halves stdio carve-out (WIRE.10), a regression test for the `NoSession` fallthrough, and a resumable continuation on `RoundsExhausted`.


MikkoParkkola:
## Independent review round: two non-Claude reviewers, one blocking finding, fixed

Both reviewers read the branch at `09f18b99` and independently raised the same
defect: `BridgeDispatcher::invoke` flattened every dispatch error into
`BridgeError::BackendFailed`, so a refusal raised *before* the request left the
gateway — no such backend, no such tool, an open circuit, a transport that never
connected — settled the idempotency key as an executed side effect. The caller was
then permanently told an operation had run that provably never did.

Fixed at `8b7a382c`:

- `BridgeError::BackendFailed` carries a `Dispatch` classification
  (`NeverReached` / `MayHaveActed`), set from `crate::Error::is_pre_dispatch`.
- The `Err` arm commits the withheld-side-effect marker only for `MayHaveActed`.
- `a_bridged_retry_refused_before_dispatch_frees_the_idempotency_key` fails with the
  old bindingless pattern and passes with the classified one — verified both ways.

Gates re-run at `8b7a382c`: `cargo test --lib gateway::meta_mcp` 657 passed / 0
failed · `cargo test --test mik_7212_mrtr7_bridge_acs` 28 passed / 0 failed ·
`cargo clippy --all-targets -- -D warnings` clean · `cargo fmt --check` clean.

Four further findings were judged out of scope for this PR — none is introduced by
it — and are filed with acceptance criteria and fail-fast checks on #569:
cancellation releasing a reservation mid-round, enforcement versus recording in
`accounted_dispatch`, a malformed `input_required` interim reading as completed, and
the bridge tests pairing an era and a capability declaration no classifier produces.

The one red check on this PR, `Public repo hygiene`, is inherited from the merge
base; #572 fixes it on `main`.

